PR #34, opened 2016-05-24
9 new model sensor type

### LINKED ISSUE #9: Sensor#create: Select Sensor Type and Sensor-ID
As journalist
I want to configure my sensors from a pre-known set of available sensor types and simple sensor IDs [e.g. "AIR21"].
So I have less trouble with configuration

roschaefer:
fixes #9 
@drjakob I called the new model ``SensorType`` which is a bit vague, to my taste. Anyways, we can add more attributes to the model if we know them from for all our customer in advance.

### DIFF
--- a/features/select_sensor_type.feature
+++ b/features/select_sensor_type.feature
@@ -1,0 +1,12 @@
+Feature: Select Sensor Type and Sensor ID
+  As journalist
+  I want to configure my sensors from a pre-known set of available sensor types and sensor IDs
+  So I have less trouble with configuration
+
+
+  Scenario: Create new Sensor and Select Sensor Type
+    Given all "Temperature" sensors measure in "Celsius"
+    When I want to create a new sensor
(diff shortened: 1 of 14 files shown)


features/select_sensor_type.feature
```diff
@@ -0,0 +1,12 @@
...
+  I want to configure my sensors from a pre-known set of available sensor types and sensor IDs
+  So I have less trouble with configuration
+
+
+  Scenario: Create new Sensor and Select Sensor Type
+    Given all "Temperature" sensors measure in "Celsius"
+    When I want to create a new sensor
+    And I select the sensor type "Temperature"
+    And I confirm the dialog
+    Then I have a new sensor in my databas
```
roschaefer: @drjakob see the new behaviour

roschaefer:
Time spent: 2 hours

Issue #18, opened 2016-05-18
Sensor.bot: Choose text components by condition


roschaefer:
As a journalist
I want text components to appear in the live report only if certain conditions hold
So the live report delivers some meaning about the recently received sensor readings

PR #41, opened 2016-05-26
18 choose text components by condition

roschaefer:
fix: #18 
@drjakob: It would be great if we had seed data for text components. Could you provide some? Then we could see these kind of changes in the live system.

CSV is just fine. I can turn it into seed data.

### DIFF
--- a/features/choose_text_components_by_condition.feature
+++ b/features/choose_text_components_by_condition.feature
@@ -1,0 +1,21 @@
+Feature: Choose Text Component by Condition
+
+  Scenario: Condition value is within range
+    Given I have these text components prepared:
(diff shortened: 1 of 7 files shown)


features/choose_text_components_by_condition.feature
```diff
@@ -0,0 +1,24 @@
...
+    When I visit the landing page
+    Then I should see:
+    """
+    I can't see a thing!
+    """
+    But I should NOT see:
+    """
+    It's nice today.
+    """
+
```
roschaefer: @drjakob two cucumber features this time, this one describes rather the issue-related functionality

drjakob:
Seed data: Bertrams Job. I'll ask him for some.

PR #57, opened 2016-06-05
56 change api

### LINKED ISSUE #56: Change API to fit with Particle-API
We cannot change how Particle sends out JSONs. For that reason we have to change our own API interface.

roschaefer:
fix #56

### DIFF
--- a/app/controllers/sensor_readings_controller.rb
+++ b/app/controllers/sensor_readings_controller.rb
@@ -49,6 +49,7 @@ class SensorReadingsController < ApplicationController
   end
 
   def sensor_reading_params
+    format_particle_api_json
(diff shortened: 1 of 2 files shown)


app/controllers/sensor_readings_controller.rb
```diff
@@ -62,4 +63,12 @@ def assign_sensor_id
...
     end
   end
+
+  # Tries to assign a missing sensor id
+  def format_particle_api_json
+    params.permit(:published_at, :core_id, :event, :data)
+    if params[:data]
+      params[:sensor_reading] = params[:data]
+    end
+  end
```
roschaefer: @kabel42 easier than expected..

PR #80, opened 2016-07-31
23 write about actual sensor value

### LINKED ISSUE #23: TextComponent#create: Write about the actual sensor value
As a journalist
I want to include the actual value of a sensor in a text component
So I can tell the reader that a particular value of e.g. 970 hPa is quite low

roschaefer:
close #23


roschaefer:
@drjakob At the moment, we don't check if the text component has a proper markup. So if you mistype the sensor name, e.g. `{ SensoXY }` you don't get an error message. Do we need a validation here?

PR #127, opened 2016-09-06
122 markup with sensor

### LINKED ISSUE #122: TextComponent: Markup with Sensor ID
Change the markup from `sensor.name` to `sensor.id`, because the ID does not have a meaning and never changes, whereas the name might change. Less errors.

roschaefer:
close #122


db/migrate/20160906224205_change_markup_from_sensor_name_to_sensor_id.rb
```diff
@@ -0,0 +1,36 @@
+class ChangeMarkupFromSensorNameToSensorId < ActiveRecord::Migration
+  def up
+    sensor_name_typos = {
```
roschaefer: @drjakob this database migration fixes several typos along with the new feature "markup via sensor id"

PR #167, opened 2016-10-03
157 end date of report

### LINKED ISSUE #157: Report: End date of report
As a reporter
I want to set an end date of my report
So text components can be aligned along a linear schedule

For convenience, the end date could also be entered in days
related: #156

roschaefer:
close #157

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -54,6 +54,6 @@ class ReportsController < ApplicationController
   end
 
   def report_params
-    params.require(:report).permit(:name, :start_date, :video, variables_attributes: [:value, :id])
+    params.require(:report).permit(:name, :start_date, :duration, :video, variables_attributes: [:value, :id])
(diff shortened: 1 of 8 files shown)


roschaefer:
@drjakob it is actually difficult to implement the change of end date via a duration (javascript, two way data binding). Just editing the end date and see the duration afterwards will do it also, I guess

drjakob:
Not sure about this one. Is this just about input?
Otherwise: The reports property "duration" is totally independent from concrete dates.

roschaefer:
The duration of an experiment is `end_date - start_date`. No?

Have a look into the cucumber feature.

drjakob:
No. First step: Define duration. (Compose & write everything). Second step: define a start_date.

roschaefer:
Ok. So `duration` is always an integer (in days)?

drjakob:
Yes. That simple.

PR #196, opened 2016-11-18
164 calibration

### LINKED ISSUE #164: Settings: Button for calibration
As member of service team
I want to have a button to start calibration of reporter.box
because easy calibration is essential for using the box.

roschaefer:
close #164

### DIFF
--- a/app/assets/javascripts/sensors.coffee
+++ b/app/assets/javascripts/sensors.coffee
@@ -8,5 +8,13 @@ ready = ->
     $('#unhide_sample_data_form').hide()
     $('#sample_data_form').show()
 
+  $('#stop_calibration').hide()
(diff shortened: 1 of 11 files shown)


roschaefer:
@kabel42 @drjakob 

This will solve #164 for now. 
The user has to start and stop the calibration. During calibration, normal sensor readings are not stored in the database but might change the `min_value` and `max_value` of the sensor .

There are two things that might not be optimal:
* If the user forgets to stop the calibration, the sensor will not be able to record normal sensor readings.
* Sensor readings during calibration are not stored in a separate table.

Please tell me if these are problems after all. I will solve them in a separate PR respectively.

roschaefer:
@drjakob @kabel42 deployed - check it out

Issue #246, opened 2017-03-07
Report Settings: Field for Chatbot-Blocks


drjakob:
As member of the service team
I want to have have an editable list of Chatfuel Block Names as output destinations ( #240) in settings
because I want send my dynamic content to multiple blocks.

PR #260, opened 2017-03-10
grow cucumber for #246

roschaefer:
@leifg grow and harvest the cucumber! ;)

increase "Estimate" with your time spent

### DIFF
--- a/features/channels/create_new_channel.feature
+++ b/features/channels/create_new_channel.feature
@@ -1,0 +1,17 @@
+@246
+Feature: Create new channels
+  As member of the service team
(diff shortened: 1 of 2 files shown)


features/channels/create_new_channel.feature
```diff
@@ -0,0 +1,17 @@
+@246
+Feature: Create new channels
```
roschaefer: If we are going to implement #246 in Sprint 8, replace all `Channel` with `Topic`

roschaefer:
outdated

Issue #243, opened 2017-03-07
TextComponents#edit: Choose one or many channels


drjakob:
As a reporter
I want to choose one or more channels for my text component
In order to associate the text component with a channel

roschaefer:
closed through #261

PR #261, opened 2017-03-10
Implement Channel and Context Data Model

roschaefer:


leifg:
Ok text components can now be published under a channel.

leifg:
@roschaefer pls review

app/views/channels/_text_components.html.haml
```diff
@@ -0,0 +1,22 @@
+- @channel.text_components.each do |text_component|
```
roschaefer: why do we have `text_components` under `view/channels` ?

app/models/text_component.rb
```diff
@@ -3,7 +3,9 @@ class TextComponent < ActiveRecord::Base
   has_and_belongs_to_many :triggers
   has_many :sensors, through: :triggers
   has_many :events, through: :triggers
+  has_and_belongs_to_many :channels
   belongs_to :report
+  belongs_to :publishing_context
```
roschaefer: what's a `publishing_context` ?

config/routes.rb
```diff
@@ -16,7 +16,10 @@
   get 'reports/current', to: 'reports#current', as: 'current_report'
   get 'reports/present/:id', to: 'reports#present', as: 'present_report'
   get 'reports/preview/:id', to: 'reports#preview', as: 'preview_report'
-  resources :reports
+
+  resources :reports do
+    resources :channels
+  end
```
roschaefer: that makes perfect sense - I have to update my own PR

app/models/text_component.rb
leifg: This is the keyword that the bot later will be triggered on. It makes way more sense to have a text component defining be under a keyword (context) than to tie the keyword to a channel.

db/migrate/20170312171020_initialize_default_contexts.rb
```diff
@@ -0,0 +1,19 @@
...
+  def up
+    PublishingContext.create!(name: "milk_quantity")
+    PublishingContext.create!(name: "milk_quality")
+    PublishingContext.create!(name: "movement")
+    PublishingContext.create!(name: "temperature")
+    PublishingContext.create!(name: "intake")
+    PublishingContext.create!(name: "birth")
+    PublishingContext.create!(name: "calf")
+    PublishingContext.create!(name: "noise")
+    PublishingContext.create!(name: "health")
```
roschaefer: I would not set up static data with migrations - we can use seeds for development and for production I would suggest to use just the normal web interface

leifg: Yeah I thought about that, but the things will likely never change. So I don't see the necessary overhead for implementing a web view.

Another solution is to implement it as enums, but then there is no chance of getting new contexts without deployment.

app/models/text_component.rb
roschaefer: I thought we could define many keywords in the ChatFuel interface? Those keywords can all be mapped to one channel

db/migrate/20170312171020_initialize_default_contexts.rb
roschaefer: I guess @drjakob would prefer a web interface. If you don't want to implement a web interface, please use a `rake task` for static data instead. I have had a lot of troubles with static data in migrations already

roschaefer: we would ssh in our production server and run a rake task once after deployment

db/seeds.rb
```diff
@@ -27,6 +27,19 @@
 Topic.find_or_create_by(name: "noise")
 Topic.find_or_create_by(name: "health")
 
+Report.all.each do |report|
```
roschaefer: This has little impact on this example, but in general it's good practice to replace`all.each` with `find_each`

app/views/channels/_text_components.html.haml
leifg: Because you need a view where you can see all text_components in a channel. This is pretty much what the cucumber feature asks for.

roschaefer:
@leifg could you add your spent hours to `estimated` hours, plz?

roschaefer:
close #243

PR #262, opened 2017-03-10
220 ux fail buttons

### LINKED ISSUE #220: TextComponent: UX Fail
"The Create Trigger" and "Create Textcomponent" buttons should be combined: "Create Text Component with New Trigger."

roschaefer:
That was an actual bug in a view
fix #220
@leifg review please

### DIFF
--- a/app/views/triggers/_trigger_fields.html.haml
+++ b/app/views/triggers/_trigger_fields.html.haml
@@ -57,7 +57,7 @@
       #events
         = f.association :events, label_method: :name_and_id, label: false
 
-- unless local_assigns[:create_trigger]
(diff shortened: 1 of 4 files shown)


app/views/triggers/_trigger_fields.html.haml
```diff
@@ -57,7 +57,7 @@
       #events
         = f.association :events, label_method: :name_and_id, label: false
 
-- unless local_assigns[:create_trigger]
+- unless local_assigns[:no_submit_button] # not a part of text component form
```
roschaefer: true. I have to use double negation, because the submit button is always there *unless* somethign is set.

I'll see if I can refactor it

PR #266, opened 2017-03-12
258 call a channel

roschaefer:
this is a piece of dummy code in order to set up a ChatFuel
@leifg I would actually like to merge this into master first, to have something we can point Chatfuel to


roschaefer:
thanks

PR #276, opened 2017-03-15
Implement chatbot responses

### LINKED ISSUE #259: Channels#show: Provide endpoint for each channel
As a developer
I want to have an endpoint for each channel which returns a message with the relevant text
In order to get the input for e.g. a chatbot

See #258

leifg:
To try it out do the following:

- Create a chatbot channel:
`Channel.create(name: "chatbot", description: "Channel for Chatbot", report: Report.current)`

- Add channel to text component.
- Add Topic "milk_quality" to text component
- Download [ngrok](https://ngrok.com/)
- Start ngrok and configure it to your local rails port: `ngrok http -subdomain=kuhbertha 3000`
- Click the "Wer bist Du?" Button in Chatbot

After introduction The text component main part should be visible

close #259

### DIFF
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -7,7 +7,6 @@ class ChatbotController < ApplicationController
       .text_components
       .where(topic: Topic.find_by(name: params[:topic]))
       .select(&:active?)
-      .tap{|tcs| puts tcs.map(&:triggers).inspect}
       .select(&:priority_raw)
       .sort_by(&:priority_raw)
       .reverse
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -1,0 +1,20 @@
+class ChatbotController < ApplicationController
+  def show
+    # TODO order by priority and maybe have a fallback
+    chatbot_channel = Channel.find_by(name: "chatbot")
+    current_report = Report.current
+    tc = TextComponent
+      .where(report: current_report)
+      .where(topic: Topic.find_by(name: params[:topic]))
+      .joins(:channels)
+      .where("channels.id = :channel_id", {channel_id: chatbot_channel.id})
+      .first
+
+    @response = {
+      messages: [
+        { text: tc.main_part.first(640)} # Messages for Facebook Messenger can only be 640 characters long. Source: https://developers.facebook.com/docs/messenger-platform/send-api-reference#request
+      ]
+    }
+    render json: @response
+  end
+end
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -6,8 +6,8 @@ class ChatbotController < ApplicationController
       .text_components
       .where(topic: Topic.find_by(name: params[:topic]))
       .select(&:active?)
-      .select(&:priority_raw)
-      .sort_by(&:priority_raw)
+      .select(&:priority_index)
+      .sort_by(&:priority_index)
       .reverse
       .first
 
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -1,6 +1,5 @@
 class ChatbotController < ApplicationController
   def show
-    # TODO order by priority and maybe have a fallback
     chatbot_channel = Channel.find_by(name: "chatbot", report: Report.current)
 
     tc = chatbot_channel
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -6,6 +6,11 @@ class ChatbotController < ApplicationController
     tc = chatbot_channel
       .text_components
       .where(topic: Topic.find_by(name: params[:topic]))
+      .select(&:active?)
+      .tap{|tcs| puts tcs.map(&:triggers).inspect}
+      .select(&:priority_raw)
+      .sort_by(&:priority_raw)
+      .reverse
       .first
 
     @response = {
--- a/app/controllers/chatbot_controller.rb
+++ b/app/controllers/chatbot_controller.rb
@@ -1,13 +1,11 @@
 class ChatbotController < ApplicationController
   def show
     # TODO order by priority and maybe have a fallback
-    chatbot_channel = Channel.find_by(name: "chatbot")
-    current_report = Report.current
-    tc = TextComponent
-      .where(report: current_report)
+    chatbot_channel = Channel.find_by(name: "chatbot", report: Report.current)
+
+    tc = chatbot_channel
+      .text_components
       .where(topic: Topic.find_by(name: params[:topic]))
-      .joins(:channels)
-      .where("channels.id = :channel_id", {channel_id: chatbot_channel.id})
       .first
 
     @response = {
--- a/app/lib/text/generator.rb
+++ b/app/lib/text/generator.rb
@@ -64,15 +64,7 @@ module Text
     end
 
     def components
-      if @components.nil?
-        @components = @report.active_sensor_story_components(@opts)
-        @components = @components.shuffle
-        @nil_priorities, @components = @components.partition {|c| c.priority.nil? }
-        @components = components.sort_by {|c| Trigger.priorities[c.priority] }
-        @components = components.reverse
-        @components = @components + @nil_priorities
-      end
-      @components
+      @compontents ||= Text::Sorter.sort(@report.text_components, @opts)
     end
   end
 end
--- a/app/lib/text/generator.rb
+++ b/app/lib/text/generator.rb
@@ -64,7 +64,7 @@ module Text
     end
 
     def components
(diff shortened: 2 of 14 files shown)


leifg:
@roschaefer please confirm if this is going in the right direction

config/routes.rb
```diff
@@ -17,6 +17,8 @@
   get 'reports/present/:id', to: 'reports#present', as: 'present_report'
   get 'reports/preview/:id', to: 'reports#preview', as: 'preview_report'
 
+  get 'chatbot/:topic', to: 'chatbot#show'
```
roschaefer: nest under `/reports/`?

app/controllers/chatbot_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def show
+    # TODO order by priority and maybe have a fallback
+    chatbot_channel = Channel.find_by(name: "chatbot")
+    current_report = Report.current
+    tc = TextComponent
+      .where(report: current_report)
+      .where(topic: Topic.find_by(name: params[:topic]))
+      .joins(:channels)
+      .where("channels.id = :channel_id", {channel_id: chatbot_channel.id})
+      .first
```
roschaefer: maybe we could even send all text components as messages

app/controllers/chatbot_controller.rb
```diff
@@ -0,0 +1,20 @@
+class ChatbotController < ApplicationController
+  def show
+    # TODO order by priority and maybe have a fallback
+    chatbot_channel = Channel.find_by(name: "chatbot")
+    current_report = Report.current
+    tc = TextComponent
+      .where(report: current_report)
+      .where(topic: Topic.find_by(name: params[:topic]))
+      .joins(:channels)
+      .where("channels.id = :channel_id", {channel_id: chatbot_channel.id})
```
roschaefer: This seems complicated to me. I would prefer `current_report.active_text_components(topic: topic)`

config/routes.rb
leifg: For now I would leave the reports out. We're implying the current report.

We should rethink the whole report structure and evaluate what it is actually good for.

leifg:
Made some changes regarding the selection for text components.

app/controllers/chatbot_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ChatbotController < ApplicationController
+  def show
+    # TODO order by priority and maybe have a fallback
+    chatbot_channel = Channel.find_by(name: "chatbot", report: Report.current)
```
roschaefer: we're expecting state in our database :cry:

app/controllers/chatbot_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    tc = chatbot_channel
+      .text_components
+      .where(topic: Topic.find_by(name: params[:topic]))
+      .select(&:active?)
+      .tap{|tcs| puts tcs.map(&:triggers).inspect}
+      .select(&:priority_raw)
+      .sort_by(&:priority_raw)
+      .reverse
+      .first
```
roschaefer: there should be tests

roschaefer: DRY out [current implementation](https://github.com/roschaefer/story.board/blob/master/app/lib/text/generator.rb#L89), please

app/models/text_component.rb
```diff
@@ -24,6 +24,10 @@ def on_time?
     result
   end
 
+  def priority_raw
```
roschaefer: namings: `priority_number` or `priority_index`

features/step_definitions/steps.rb
```diff
@@ -727,13 +727,23 @@
...
 Given(/^we created a text component for it that is active right now$/) do
-  main_part = "If you think there is good in everybody, you haven't met everybody."
-  create(:text_component, :active, main_part: main_part, channels: [@channel])
+  create(
+    :text_component,
+    :active,
+    main_part: "Got milk?",
+    channels: [@channel],
+    topic: @topic,
+    triggers: [create(:trigger, priority: "high")]
```
roschaefer: remove `triggers: [...]` - don't need that to make the test pass

app/controllers/chatbot_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ChatbotController < ApplicationController
+  def show
+    # TODO order by priority and maybe have a fallback
+    chatbot_channel = Channel.find_by(name: "chatbot", report: Report.current)
+
+    tc = chatbot_channel
+      .text_components
+      .where(topic: Topic.find_by(name: params[:topic]))
+      .select(&:active?)
+      .tap{|tcs| puts tcs.map(&:triggers).inspect}
```
roschaefer: remove debugging output

app/controllers/chatbot_controller.rb
roschaefer: btw: this will fail for `priority = nil`

app/controllers/chatbot_controller.rb
leifg: So?

features/step_definitions/steps.rb
leifg: Yes you do

app/lib/text/generator.rb
```diff
@@ -91,7 +91,7 @@ def components
         @components = @report.active_text_components(@opts)
         @components = components.shuffle
         @nil_priorities, @components = components.partition {|c| c.priority.nil? }
-        @components = components.sort_by {|c| Trigger.priorities[c.priority] }
+        @components = components.sort_by(&:priority_index)
```
roschaefer: After looking at your second PR, I can see another shared responsibility: 3. Substitution of markup. If you decompose and reuse that part, you can be quite sure that the chatfuel controller can also deliver messages like: `It is 32°C right now.` There are tests for this functionality already, so you wouldn't need to test that again.

features/step_definitions/steps.rb
roschaefer: By looking at the [code on Github](https://github.com/roschaefer/story.board/blob/master/app/models/text_component.rb#L28), you can see that the priority of a `TextComponent` is the highest priority of it's `triggers`. No triggers means `select(&:priority_index)` will return an empty array.

roschaefer: It would have been awkward if there were a factory with a trait called `:active` that does not return an active text component.

app/lib/text/generator.rb
leifg: The rendering is now done in a separate classe. Will look at selecting the components tonigh

leifg:
@roschaefer pls review again.

We should merge #283 before merging this, so we can use the `chatbot` channel scope

features/step_definitions/steps.rb
roschaefer: After you use `Text::Sorter` in Chatfuel Controller, you don't need `triggers` to activate a text component.

spec/controllers/chatfuel_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  context "matching text component" do
+    let(:main_part) { "Main Part" }
+    let!(:text_component) do
+      create(
+        :text_component,
+        report: report,
+        topic: topic,
+        channels: [chatbot_channel],
+        triggers: [create(:trigger, priority: "high")],
```
roschaefer: same here

spec/text/renderer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+  let(:report_name) { 'ReportName' }
+  let(:intention)      { :real }
+  subject { described_class.new(text_component: text_component, opts: {intention: intention}) }
+
+  describe '#render' do
+    let(:part) { :main_part }
+    subject { super().render(part) }
+
+    context 'given one text_component' do
+      let(:text_components) { [text_component] }
```
roschaefer: `Text::Renderer` renders one text component at a time? Remove `let(:text_components) { .. }`

spec/text/sorter_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      let(:text_components) do
+        [
+          create(:text_component, heading: 'Text component 2', triggers: []),
+          create(:text_component, heading: 'Text component 1', triggers: [create(:trigger, priority: :medium)]),
+        ]
+      end
+
+      it 'returns text sorted by trigger' do
+        expect(subject.sort(text_components, opts)).to eq(expected_result)
+      end
```
roschaefer: You *can* actually test for a random selection in case of a tie. Just call `sort` several times and check if the results are different The previous test for randomness is gone.

features/step_definitions/steps.rb
leifg: Nice! Just removed it

leifg:
Incorporated latest CR comments

PR #279, opened 2017-03-17
277 double report input fields

### LINKED ISSUE #277: TextComponent: Double report input fields without default
selection_112

roschaefer:
close #277 
@leifg review plz

### DIFF
--- a/app/views/text_components/_form.html.haml
+++ b/app/views/text_components/_form.html.haml
@@ -8,7 +8,7 @@
         #triggers
           - if local_assigns[:create_trigger]
             = f.simple_fields_for :triggers do |trigger|
-              = render 'triggers/trigger_fields', f: trigger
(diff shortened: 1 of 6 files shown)


app/views/text_components/_form.html.haml
```diff
@@ -8,7 +8,7 @@
         #triggers
           - if local_assigns[:create_trigger]
             = f.simple_fields_for :triggers do |trigger|
-              = render 'triggers/trigger_fields', f: trigger
+              = render 'triggers/trigger_fields', f: trigger, is_sub_form: true
```
roschaefer: I'm using conditionals here again, because I have to change the view at two different points - is there a cleaner solution?

leifg: Conditionals are fine.

PR #283, opened 2017-03-17
Remove text components from sensorstory if they are not in the right channel

leifg:
this is a pretty messy solution. I am currently not aware of a better
solution though.

### DIFF
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -1,4 +1,8 @@
 class Channel < ActiveRecord::Base
   has_and_belongs_to_many :text_components
   belongs_to :report
+
+  def self.default(report)
+    find_by(name: "sensorstory", report: report)
+  end
 end
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -2,7 +2,10 @@ class Channel < ActiveRecord::Base
   has_and_belongs_to_many :text_components
   belongs_to :report
 
+  scope :chatbot,     -> { find_by(report: Report.current, name: "chatbot") }
+  scope :sensorstory, -> { find_by(report: Report.current, name: "sensorstory") }
+
   def self.default(report)
-    find_by(name: "sensorstory", report: report)
+    Channel.sensorstory
   end
 end
--- a/app/models/channel.rb
+++ b/app/models/channel.rb
@@ -5,6 +5,8 @@ class Channel < ActiveRecord::Base
   scope :chatbot,     -> { find_by(report: Report.current, name: "chatbot") }
   scope :sensorstory, -> { find_by(report: Report.current, name: "sensorstory") }
 
+  validates :report, presence: true
+
   def self.default(report)
     Channel.sensorstory
   end
--- a/app/models/text_component.rb
+++ b/app/models/text_component.rb
@@ -9,7 +9,7 @@ class TextComponent < ActiveRecord::Base
 
   accepts_nested_attributes_for :triggers
 
-  before_create :assign_default_channel
+  validates :channels, presence: true
 
   def active?(opts={})
     on_time? && triggers.all? {|t| t.active?(opts) }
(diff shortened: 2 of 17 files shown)


app/models/text_component.rb
```diff
@@ -9,6 +9,8 @@ class TextComponent < ActiveRecord::Base
 
   accepts_nested_attributes_for :triggers
 
+  before_create :assign_default_channel
```
roschaefer: Please, don't auto-assign a channel in `before_create`. I would use `before_validation` but actually I wouldn't auto-assign a channel in this particular case at all. This will be called whenever we create a text component! It will affect all our tests. Imagine you do `create(:text_component, channel: nil)` and suddenly the component is assigned to a channel.

I suggest to use a validation `validates :channel, presence: true` (+test) and to set a default channel in the view, e.g. `f.association :channel, allow_blank: false`. (+test, either a view test or a cucumber test).

app/models/channel.rb
```diff
@@ -1,4 +1,8 @@
 class Channel < ActiveRecord::Base
   has_and_belongs_to_many :text_components
   belongs_to :report
+
+  def self.default(report)
+    find_by(name: "sensorstory", report: report)
```
roschaefer: If we expect certain records in our database, we should use a constant like `Channel::SENSORSTORY` and `Channel::CHATFUEL`. I believe that scopes are quite handy for that, sth. like `SENSORSTORY = ~> { find_by(name: 'sensorstory') }`

leifg: That's a good idea, have not thought of that.

leifg: Although the constant will not work. The constant will be initialized when the application starts. If for any reason the database changes, this changes will go unnoticed (hello breaking tests).

I will introduce a scope:

```
  scope :chatbot,     -> { find_by(report: Report.current, name: "chatbot") }
  scope :sensorstory, -> { find_by(report: Report.current, name: "sensorstory") }
```

which can then be used by:

```
Channel.chatbot
```

app/models/text_component.rb
leifg: I changed the before_validation to expect channels. This breaks ~50 tests of different kinds. Changing the default text_component factory will only fix part of them.

Can you have a look?

spec/text/generator_spec.rb
```diff
@@ -1,146 +1,151 @@
...
-  let(:intention)      { :real }
+  let!(:sensorstory_channel)  { create(:channel, name: "sensorstory", report: report) }
+  let(:report)                { create(:report) }
+  let(:intention)             { :real }
   subject { described_class.new(report: report, opts: {intention: intention}) }
 
-  describe '#choose_heading' do
-    subject { super().choose_heading }
-    let(:report) { create(:report, text_components: text_components) }
+  context 'for text components with the same report and the same channel' do
```
roschaefer: We are testing this situation and only this situation. We cannot say what happens for any other combination. If we really want to stick to static data in the database, let's make `Channel` a static read-only model and generate the two entries with a migration. I dislike seed data in migrations, but actually that wouldn't be seed data, because we expect it in our tests as well.

spec/models/channel_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe Channel, type: :model do
+  describe 'factory' do
+    subject { build(:channel) }
+    it { is_expected.to be_valid }
+  end
+
+  describe 'channel without report' do
+    subject { build(:channel, report: nil) }
```
roschaefer: There is a view spec failing for a good reason

spec/factories/channels.rb
```diff
@@ -1,5 +1,11 @@
 FactoryGirl.define do
   factory :channel do
     name "sensorstory"
+
+    initialize_with do
+      Channel.find_or_create_by(name: name) do |c|
+        c.report = build(:report)
+      end
+    end
```
roschaefer: That will affect all our tests, right? Does it create reports randomly?

PR #299, opened 2017-03-30
added missing channel data to seeds.rb

matthib:
Added missing channel data to seeds.rb, so the project can find the necessary channels and work as expected locally,


roschaefer:
excellent!
In the future this should be done with a migration. Purpose of seed data is to have some data for development. Seed data, in contrast to migrations, should not be essential for local setup.

PR #319, opened 2017-04-23
206 report read more button

### LINKED ISSUE #206: Report#show: Read more button
As a reader who is sometimes more, sometimes less interested
I want small reports with a "Read more" button
To keep reports brief with the option to read on if I am really interested

roschaefer:
@drjakob this is a draft to close #206 

I don't think this is finished yet, can you give me some feedback on the following?
* Question/answers show up only for the live report
* They show up below the issue
* Questions are simple buttons, answers are just paragraphs
* All questions are visible at once

### DIFF
--- a/app/assets/javascripts/report.coffee
+++ b/app/assets/javascripts/report.coffee
@@ -1,3 +1,0 @@
-# Place all the behaviors and hooks related to the matching controller here.
-# All this logic will automatically be available in application.js.
-# You can use CoffeeScript in this file: http://coffeescript.org/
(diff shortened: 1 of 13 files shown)


drjakob:
Question/answers show up only for the live report –> Not in PREVIEW?
They show up INSIDE THE TEXT, after each main paragraph
Questions are simple buttons, answers are just paragraphs –> YES
All questions are visible at once –> No. Only the first Question is visible. If the answer is shown, the next question is shown.

PR #322, opened 2017-04-27
Chatfuel questions and answers

### LINKED ISSUE #248: Chatfuel: Read more question
As a reader using the Facebook Messenger
I want to be asked, if i want to continue reading
In order not to be overwhelmed by the amount of text

matthib:
This is the pull request for issue #248.

If a text component contains question and answers, send button with question to chatfuel. As long as there are more questions, keep sending the answer of the last question and a button for the new question.

The formatting is not always right in this commit (indentation with 4 chars vs. 2 chars), this needs to be addressed later.

---

close #248

### DIFF
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -2,88 +2,63 @@ class ChatfuelController < ApplicationController
   def show
     chatbot_channel = Channel.chatbot
 
-    tc = Text::Sorter.sort(chatbot_channel.text_components, {}).first
+    @text_component = Text::Sorter.sort(chatbot_channel.text_components, {}).first
 
-    if tc
-      text = Text::Renderer.new(text_component: tc).render(:main_part)
-
-      qa = tc.question_answers
-
-      #Test API Request: http://localhost:3000/chatfuel/1
-      if qa.exists?(1)
-          sendJsonResponse(text, 1, qa.find(1).question)
-      else
-          sendJsonResponse(text)
-      end
-
+    if @text_component
+      @next_question_answer = @text_component.question_answers.first
+      render json: json_response
     else
-     render json: {}, status: 404
+      render json: {}, status: 404
     end
   end
 
-  def showQuestionAndAnswer
-      # request for qa for a specific text component
-      # send answer as json response to chatfuel
-      # if more questions and answers available, append button for next answer to json response
-
-      chatbot_channel = Channel.chatbot
-
-      tc = Text::Sorter.sort(chatbot_channel.text_components, {}).first
-
-      if tc
-          qaId = params[:id]
-
-          if tc.question_answers.exists?(qaId)
-              qa = tc.question_answers.find(qaId)
-
-              if tc.question_answers.exists?(qaId.to_i + 1)
-                  # send answer + button for next question
-                  sendJsonResponse(qa.answer, qaId.to_i + 1, tc.question_answers.find(qaId.to_i + 1).question)
-              else
-                  # send only answer for this question
-                  sendJsonResponse(qa.answer)
-              end
-          else
-              render json: {}, status: 404
-          end
-
-      else
-          render json: {}, status: 404
-      end
+  def answer_to_question
+    @text_component = TextComponent.includes(:question_answers).find(params[:text_component_id])
+    index = params[:index].to_i
+    @question_answer = @text_component.question_answers[index]
+    @next_question_answer = nil # TODO: implement
+    render json: json_response
   end
 
-  def sendJsonResponse(text, qaId = 0, questionText = '')
-      if qaId.to_i == 0
-          @response = {
-            messages: [
-              { text: text.first(640)} # Messages for Facebook Messenger can only be 640 characters long. Source: https://developers.facebook.com/docs/messenger-platform/send-api-reference#request
-            ]
-          }
-      else
-          # send json response with text + qa button
-          # todo: create url for next question from question id
-          @response = {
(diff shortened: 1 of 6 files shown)


roschaefer:
rename branch `git branch -m 248-chatfuel-questions-and-answers`

roschaefer:
close #248 is the right markup to close an issue within the description of a pull request

roschaefer:
create at least one cucumber test (I will help)

roschaefer:
Ok, deleting the branch remotely will close this PR. Just prefix your branches with the issue number for the future.

app/controllers/chatfuel_controller.rb
```diff
@@ -7,14 +7,83 @@ def show
...
+          else
+              render json: {}, status: 404
+          end
+
+      else
+          render json: {}, status: 404
+      end
+  end
+
+  def sendJsonResponse(text, qaId = 0, questionText = '')
```
roschaefer: Add at least a unit test for `#sendJsonResponse` and a controller test for `#showQuestionAndAnswer` (it's a controller action).

app/controllers/chatfuel_controller.rb
```diff
@@ -7,14 +7,83 @@ def show
...
-        messages: [
-          { text: text.first(640)} # Messages for Facebook Messenger can only be 640 characters long. Source: https://developers.facebook.com/docs/messenger-platform/send-api-reference#request
-        ]
-      }
-      render json: @response
+      qa = tc.question_answers
+
+      #Test API Request: http://localhost:3000/chatfuel/1
+      if qa.exists?(1)
+          sendJsonResponse(text, 1, qa.find(1).question)
```
roschaefer: Why don't you dispatch with two different methods?

app/controllers/chatfuel_controller.rb
```diff
@@ -7,14 +7,83 @@ def show
...
-      @response = {
-        messages: [
-          { text: text.first(640)} # Messages for Facebook Messenger can only be 640 characters long. Source: https://developers.facebook.com/docs/messenger-platform/send-api-reference#request
-        ]
-      }
-      render json: @response
+      qa = tc.question_answers
+
+      #Test API Request: http://localhost:3000/chatfuel/1
+      if qa.exists?(1)
```
roschaefer: I never use `::exists?(id)`. According to [ApiDoc](https://apidock.com/rails/ActiveRecord/Base/exists%3F/class), this will find a QuestionAnswer with `id == 1`. Why do you want to find a particular QuestionAnswer record?

roschaefer:
@matthib I rebased this on top of master `git rebase master -i`. Please do `git fetch` followed by `git reset origin/chatfuel-questions-and-answers --hard`

app/controllers/chatfuel_controller.rb
matthib: Do you mean the method sendJsonResponse() why I use it in both cases of the if statement?

roschaefer: yup

app/controllers/chatfuel_controller.rb
roschaefer: Go and check if `exists(1)` really behaves that way. I guess it just checks for a record with id 1.

app/controllers/chatfuel_controller.rb
matthib: I guess I could use two methods, I just tried to combine the action of sending an json response into one method. But the two responses differ a lot, so probably they could have their own method each.

app/controllers/chatfuel_controller.rb
matthib: "Returns true if a record exists in the table that matches the id or conditions given, or false otherwise."
-> So qa.exists?(1) will return true if a QuestionAnswer with the id 1 exists.

I think i see what you mean, the first QuestionAnswer record for a certain topic doesn't need to have the id 1. 

If used if qa.first instead, would that be the right implementation?

roschaefer: `qa.first` !

config/routes.rb
```diff
@@ -18,6 +18,9 @@
   get 'reports/preview/:id', to: 'reports#preview', as: 'preview_report'
 
   get 'chatfuel/:topic', to: 'chatfuel#show'
+  # route for chatfuel questions and answers
+  get 'chatfuel/:topic/qa/:id', to: 'chatfuel#showQuestionAndAnswer', as: 'chatfuel_qa'
```
roschaefer: In ruby land it is convention to use `snake_case` for methods and `CamelCase` for constants, e.g. class names. Constants require a leading upcase letter.

roschaefer:
hey @matthib I did a heavy refactoring

Just for the record: Variables starting with `@` are instance variables and in controllers they can be used to share state among private methods

features/channels/question_answer.feature
```diff
@@ -24,7 +24,7 @@ Feature: Read more question
               "text": "The main part of the text component will be displayed here.",
               "buttons": [
                 {
-                  "url": "http://localhost:3000/chatfuel/text_components/1/answer_to_question/1",
+                  "url": "http://example.org/chatfuel/text_components/178/answer_to_question/1",
```
roschaefer: change that to
```ruby
"http://example.org/chatfuel/text_components/#{@text_component.id}/answer_to_question/1"
```

roschaefer:
@matthib please check out the refactoring and the tests. I would merge this PR as soon our CI server is green.

PR #363, opened 2017-05-20
Added authorization validation for writing functions

matthib:
I added the devise (our rails authentication solution) helper function authenticate_user! for controller functions that write / manipulate data.

This only protects the functions from being called, in the template the link to the functions (i.e. 'Edit' or 'Destory') is still visible. Maybe this should also be changed in the future.

### DIFF
--- a/app/controllers/triggers_controller.rb
+++ b/app/controllers/triggers_controller.rb
@@ -1,5 +1,6 @@
 class TriggersController < ApplicationController
   before_action :set_trigger, only: [:show, :edit, :update, :destroy]
+  #before_action :authenticate_user!, only: [:new, :edit, :create, :update, :destroy]
 
   def index
     @triggers = Trigger.all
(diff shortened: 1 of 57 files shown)


app/controllers/triggers_controller.rb
```diff
@@ -1,5 +1,6 @@
 class TriggersController < ApplicationController
   before_action :set_trigger, only: [:show, :edit, :update, :destroy]
+  before_action :authenticate_user!, only: [:new, :edit, :create, :update, :destroy] unless Rails.env.test?
```
roschaefer: `unless Rails.env.test` is a big no-no :crying_cat_face:

roschaefer: You have to change the respective tests. Add `Given I am logged in` where necessary.

app/views/actuators/index.html.haml
```diff
@@ -6,18 +6,20 @@
       %th Name
       %th Port
       %th
-      %th
-      %th
+      - if user_signed_in?
+        %th
+        %th
```
roschaefer: remove this and add empty `%td` if the user is not signed in

roschaefer:
@matthib finally!! :tada: :sweat:

PR #366, opened 2017-05-22
311 use published text components only

### LINKED ISSUE #311: Text.components: Use only published components
As a reporter (Björn)
I want the status (described in #310) that only text.components with status "published" are used to generate the report
in order to work on the running system.

tillprochaska:
close #311


app/models/text_component.rb
```diff
@@ -17,6 +17,8 @@ class TextComponent < ActiveRecord::Base
   delegate :name, to: :topic, prefix: true, allow_nil: true
   delegate :email, to: :assignee, prefix: true, allow_nil: true
 
+  enum publication_status: [ :draft, :fact_checked, :published ]
```
roschaefer: if you have a default value in the migration, you should make this a Hash and define the value `0`

features/text_components/publication_status.feature
```diff
@@ -0,0 +1,27 @@
+@311
+Feature: Only published text components should used to generate a report
+  As a reporter
+  I want to have a flag for text.components (draft, fact-checked, published) and only published text components should be included in the report
+  In order to add new text components without affecting the result of the live report
```
roschaefer: Try to explain the value: Why does sb. want to add a text component without affecting the report? Maybe to show it to other team members?

features/text_components/publication_status.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  Background:
+    Given there is a sensor live report
+  
+  Scenario:
```
roschaefer: This scenario is perfectly valid. But how about describing that the journalist *changes* the `publication_status` and then suddenly the text component is visible in the text? Is it better, ie. more descriptive?

spec/factories/text_components.rb
```diff
@@ -11,6 +11,8 @@
 
     channels { [Channel.sensorstory] }
 
+    publication_status 2
```
roschaefer: you can also provide a symbol here: `:published`

tillprochaska:
Hi @roschaefer! Thanks for the review, I’ll have a closer look at this later.

One more question: Would you add further unit or integration tests? I’m somehow afraid of adding tests that are unnecessary (like the one I finally removed…).

roschaefer:
A test is necessary if and only if you break sth. and no test fails after you introduced the bug. Having that said, too many tests are possibly better than too few. The only reasons for not writing a test is to avoid extra time on the build server (no issue for us, builds take < 5 min) or to avoid complicated refactoring if your tests are too closely linked with your implementation. Your implementation is unlikely to be under heavy refactoring. So go ahead!

app/views/text_components/_text_component_row.html.haml
```diff
@@ -3,6 +3,9 @@
   %td= text_component.from_day
   %td= text_component.to_day
   %td
+    - if status_class_mapping = {'draft' => 'default', 'fact_checked' => 'warning', 'published' => 'success'} and status_class_mapping.key? text_component.publication_status
```
roschaefer: don't use `and` for conditionals but for control flow: http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/

roschaefer: define `css_class_mapping` in the model (or in a decorator :smile_cat: )

tillprochaska: Okay, I can understand that this wasn’t the most elegant way to solve this (and I have moved this to a model method instead to avoid cluttered views). But let me ask one more question just to make sure I understand this correctly… ;)

Is there anything wrong about using `and` in this specific example (other than the reason stated above)? I am actually using it to chain those two expressions, right?

roschaefer: `and` has lower precedence than `&&`. It's a possible cause for errors when used in a `if` clause.

app/models/text_component.rb
```diff
@@ -40,4 +42,21 @@ def priority
     most_important_trigger = triggers.sort_by {|t| Trigger.priorities[t.priority] }.reverse.first
     most_important_trigger && most_important_trigger.priority
   end
+
+  def css_status_class
```
roschaefer: this would be a great method to test in a Unit test ;)

features/step_definitions/steps.rb
```diff
@@ -370,6 +370,14 @@
   create(:text_component, main_part: main_part, report: Report.current)
 end
 
+Given(/^there is an unpublished text component with the following main part:$/) do |main_part|
+  @text_component = create(:text_component, main_part: main_part, report: Report.current, publication_status: :draft)
+end
+
+Given(/I change the text component's publication status to "([^"]*)"$/) do |status|
+  @text_component.update! publication_status: status
```
roschaefer: I would expect that `I change sth.` refers to a change in the user interface and not a low-level-change in the database (I am not a hacker!). I assume, the user interface is not yet finished?

PR #367, opened 2017-05-25
Button with link to block after last question and answer

### LINKED ISSUE #340: Chatfuel: CALL a block when question_answers are empty
As a user of the facebook messenger
I want to click on a button and get back to where I came from when the list of question-answers is empty
in order to continue my talk in chatfuel

matthib:
With the last answer of a question & answer list a button is sent linking to a block (name pattern for the block is cont_+[topic]) in order to continue with the conversation.

---
Close #340

### DIFF
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -53,6 +53,26 @@ class ChatfuelController < ApplicationController
           }
         }
       ]
+    elsif @question_answer && !@next_question_answer
+      topic = @text_component.topic
+      messages =  [
+          {
+          attachment: {
+            payload: {
+              template_type: "button",
+              text: "#{content}",
+              buttons: [
+                {
+                  type: "show_block",
+                  block_name: "cont_" + topic.name,
+                  title: "Zurück",
+                }
+            ]
+            },
+            type: "template"
+          }
+        }
+      ]
     else
       messages =  [
         { text: content.first(640)} # Messages for Facebook Messenger can only be 640 characters long. Source: https://developers.facebook.com/docs/messenger-platform/send-api-reference#request
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -14,9 +14,10 @@ class ChatfuelController < ApplicationController
 
   def answer_to_question
     @text_component = TextComponent.includes(:question_answers).find(params[:text_component_id])
+    @topic = @text_component.topic
     index = params[:index].to_i - 1
     @question_answer = @text_component.question_answers[index]
-    if @question_answer
+    if @question_answer && @topic
       @next_question_answer = @text_component.question_answers[index + 1]
       render json: json_response
     else
@@ -54,7 +55,6 @@ class ChatfuelController < ApplicationController
         }
       ]
     elsif @question_answer && !@next_question_answer
-      topic = @text_component.topic
       messages =  [
           {
           attachment: {
@@ -64,7 +64,7 @@ class ChatfuelController < ApplicationController
               buttons: [
                 {
                   type: "show_block",
-                  block_name: "cont_" + topic.name,
+                  block_name: "cont_" + @topic.name,
                   title: "Zurück",
                 }
             ]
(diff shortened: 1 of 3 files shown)


app/controllers/chatfuel_controller.rb
```diff
@@ -53,6 +53,26 @@ def json_response
...
+                  block_name: "cont_" + topic.name,
+                  title: "Zurück",
+                }
+            ]
+            },
+            type: "template"
+          }
+        }
+      ]
     else
```
roschaefer: this else block IS what shall happen if there is no question answer anymore - use it

app/controllers/chatfuel_controller.rb
```diff
@@ -53,6 +53,26 @@ def json_response
...
+      messages =  [
+          {
+          attachment: {
+            payload: {
+              template_type: "button",
+              text: "#{content}",
+              buttons: [
+                {
+                  type: "show_block",
+                  block_name: "cont_" + topic.name,
```
roschaefer: what happens if `text_component.topic` is nil? Could you please write a test for that? (rspec test)

matthib: I would return a 404 error, if no topic exists. Is it actually allowed for a text_component to not have a topic?

app/controllers/chatfuel_controller.rb
matthib: I didn't use the else block because I think it is also used by the show function (which also calls json_response), if only a chatfuel topic is requested, without question_answers. So I figured I had to use a block only for the use case that there are question_answers and it is the last question_answer.

roschaefer: Oh that could be! When does the show reach the `else` block exactly? Could you write a test that reaches the ordinary `else` block?

roschaefer: I cannot see a `validates :topic, presence: true` in https://github.com/roschaefer/story.board/blob/master/app/models/text_component.rb so I guess it's possible for a text component not to have a topic. It's good that you ask btw. Some developers add `nil` checks when it can never happen - that' what I call `Angstcode` :fearful:

matthib: I added respec test for the route /chatfuel/:topic, which will reach the else block if no errors occur to display the text component's text.

roschaefer: For your information, the next time we update, all our `belongs_to` associations will be required by default: http://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html

spec/requests/chatfuel_spec.rb
```diff
@@ -7,6 +7,48 @@
...
+            channels: [chatbot_channel],
+            main_part: "The main part"
+          )
+        end
+
+        it { is_expected.to have_http_status(:ok) }
+
+        it 'should show text component text' do
+          json_response = JSON.parse(subject.body)
+          expect(json_response['messages'][0]['text']).to eq 'The main part'
```
roschaefer: So we reach the else block if and only if we call the topic the first time and there is only one text component and this text component has no `question_answers`?

matthib: The else block is reached, if the route /chatfuel/:topic is called and there is at least one text component and it has no questions and answers. Then just the main part of that text component is rendered in the else block.
I just realized that when calling the route /chatfuel/:topic the :topic part is irrelevant, because the :topic is never accessed, but the first text_component from the chatbot channel is accessed.

roschaefer: Always break sth. and check if a test fails. That's called "Mutation testing" btw.

roschaefer: So is this PR now following to the requested behaviour? Can we merge this? I'm still not sure, but we can also simply merge this and then come back to @drjakob

roschaefer: Just tell me and I click on "approve"

drjakob: I think its like requested.

PR #383, opened 2017-05-31
358 render any markup

### LINKED ISSUE #356: TextRenderer: Write about unconnected sensor
change the behaviour of https://github.com/roschaefer/story.board/blob/master/spec/text/renderer_spec.rb#L98 to the opposite

### LINKED ISSUE #358: BUG inserted markup by menu will not render
The inserted markup in text.component create via new menu is not rendered correctly. 
– Sensor-Markup should Output the Sensor value, e.g. 37° C
– Event-Markup should Output the Date and Time in German Format (important!), eg. 27. Mai, 9:17 Uhr.

tillprochaska:
close #358 
close #356

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -6,10 +6,11 @@ class EventDecorator
 
   def date
     date = @event.happened_at
-    unless date.nil?
+    
+    if date
       date.in_time_zone(Report::TIME_ZONE).strftime(Report::DATE_FORMAT)
     else
-      "NaN"
+      "-- missing event data --"
     end
   end
 
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -6,7 +6,11 @@ class EventDecorator
 
   def date
     date = @event.happened_at
-    date.in_time_zone(Report::TIME_ZONE).strftime(Report::DATE_FORMAT)
+    unless date.nil?
+      date.in_time_zone(Report::TIME_ZONE).strftime(Report::DATE_FORMAT)
+    else
+      "NaN"
+    end
   end
 
   def method_missing(m, *args, &block)
--- a/app/decorators/sensor_decorator.rb
+++ b/app/decorators/sensor_decorator.rb
@@ -6,9 +6,13 @@ class SensorDecorator
 
   def last_value(opts={})
     r = @sensor.last_reading(opts)
-    v = r.calibrated_value
     u = @sensor.sensor_type.unit
-    "#{format("%.1f", v)}#{u}"
+    unless r.nil?
+      v = r.calibrated_value
+      "#{format("%.1f", v)} #{u}"
+    else
+      "NaN #{u}"
+    end
   end
 
   def method_missing(m, *args, &block)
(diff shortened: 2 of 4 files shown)


app/decorators/sensor_decorator.rb
```diff
@@ -6,9 +6,13 @@ def initialize(sensor)
 
   def last_value(opts={})
     r = @sensor.last_reading(opts)
-    v = r.calibrated_value
     u = @sensor.sensor_type.unit
-    "#{format("%.1f", v)}#{u}"
+    unless r.nil?
```
roschaefer: better:
```ruby
if reading
  "#{format("%.1f", v)}#{u}"
else
  "-- missing sensory data -- "
end

Comment (roschaefer) [inline app/decorators/sensor_decorator.rb]: In languages like ruby, everything that is neither `false` or `nil` evaluates to true. So to check for nil you can just write `if reading`.

File: app/decorators/event_decorator.rb [hunk 2]
```diff
@@ -6,7 +6,11 @@ def initialize(event)
 
   def date
     date = @event.happened_at
-    date.in_time_zone(Report::TIME_ZONE).strftime(Report::DATE_FORMAT)
+    unless date.nil?
+      date.in_time_zone(Report::TIME_ZONE).strftime(Report::DATE_FORMAT)
+    else
+      "NaN"
```
roschaefer: Don't use abbreviations like "NaN" (Not-a-number) people without a tech background don't understand that.

app/decorators/sensor_decorator.rb
```diff
@@ -6,9 +6,13 @@ def initialize(sensor)
...
+      v = r.calibrated_value
+      "#{format("%.1f", v)}#{u}"
+    else
+      "NaN #{u}"
```
roschaefer: "NaN temperature"? just use a common placeholder

app/lib/text/renderer.rb
```diff
@@ -17,14 +17,27 @@ def render_string(template)
       else
         rendered = template
         rendered = render_report(rendered)
-        @text_component.sensors.each do |sensor|
-          s = SensorDecorator.new(sensor)
-          rendered.gsub!(/({\s*#{ Regexp.quote("value(#{s.id})") }\s*})/, s.last_value(@opts))
+        
+        rendered.scan(/{\s*value\(\s*(\d+)\s*\)\s*}/).each do |sensor_id|
```
roschaefer: could you add a test for that? probably you can remove `\s*`.

roschaefer: Maybe you can remove the Regex altogether and replace it with Ruby's format strings: https://ruby-doc.org/core-2.2.0/String.html#method-i-25

app/decorators/sensor_decorator.rb
roschaefer: In what scenario do we need to check for `nil` actually? Could you write an Rspec test for that?

tillprochaska: > In what scenario do we need to check for nil actually? Could you write an Rspec test for that?

This would throw an exception if a sensor exists but there aren’t any associated sensor readings. I’ve extended the existing rspec test for the sensor decorator to handle this case.

app/lib/text/renderer.rb
tillprochaska: I have undone most of this change as it hasn’t been well thought out. There are a couple of test for the text render for all possible scenarios. Are you missing anything specific?

PR #389, opened 2017-06-05
365 call smaxtecs api

### LINKED ISSUE #365: CronJob: Call the smaxtecs API
As a reporter
I want to get the temperature from inside sensor as sensor value
in order to report better about the cows. (As epic #334)

matthib:
This branch calls the Smaxtec Test API with the route /smaxtec_api/temperature and returns the temperature of a test cow as a JSON.

It does not write the temperature data to Sensor Readings yet.

What is also missing is a good way to store the Smaxtec API login information without having it in our source code on github, which also works for the build server.

This PR does pretty much what is expected in #365, authenticating with the Smaxtec API and fetching the last temperature from a cow. The next step (writing the sensor data to the corresponding sensor) will be part of the related issues of the next milestone.

Close #365


spec/requests/smaxtec_api_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe 'Smaxtec API', type: :request do
+
+  describe 'GET' do
+    let(:url) { '/smaxtec_api/temperature' }
```
roschaefer: where is the test?

config/routes.rb
```diff
@@ -22,6 +22,8 @@
   # route for chatfuel questions and answers
   get 'chatfuel/text_components/:text_component_id/answer_to_question/:index', to: 'chatfuel#answer_to_question', as: 'answer_to_question'
 
+  get '/smaxtec_api/temperature', to: 'smaxtec_api#get_temperature'
```
roschaefer: don't add a route, calling the smaxtecs api should not be triggered with a request to the story.board but rather by a cronjob

roschaefer: see this file for suggestions: https://github.com/roschaefer/story.board/blob/master/config/schedule.rb

roschaefer: and https://github.com/roschaefer/story.board/blob/master/lib/tasks/archive.rake

roschaefer: and this gem: https://github.com/javan/whenever

spec/lib/smaxtec_api_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+VCR.configure do |config|
+  config.cassette_library_dir = 'fixtures/vcr_cassettes'
+  config.hook_into :webmock
+end
+
+RSpec.describe SmaxtecApi do
+    let(:smaxtec_api) { SmaxtecApi.new() }
+
+  describe '.get_temeperature' do
```
roschaefer: typo

app/lib/smaxtec_api.rb
```diff
@@ -0,0 +1,67 @@
+class SmaxtecApi < ApplicationController
```
roschaefer: just remove `< ApplicationController` I don't see any controller specific methods, so this should still work

Apart from that this PR looks ready to merge

PR #401, opened 2017-06-10
385 Chatfuel api request return text_component based on topic.name

### LINKED ISSUE #385: BUG Chatfuel API request
Chatfuel API request always returns text component for topic "milk_quantity". 
e.g. a GET request from Chatfuel to https://redaktion.superkuehe.de/chatfuel/noise should (at the moment) always return text component named "TESTBAUSTEIN Lärm" with topic "noise" but returns the text component with topic "milk_quantity".

matthib:
Fixed the bug where the Chatfuel API request would always return the same text component, not looking for the requested topic.

Now the topic request looks for the topic with the corresponding topic name and returns the first text component of this topic.

---

Close #385

### DIFF
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -3,7 +3,8 @@ class ChatfuelController < ApplicationController
     @topic = Topic.find_by(name: params[:topic])
 
     if @topic
-      @text_component = Text::Sorter.sort(@topic.text_components, {}).first
+      text_components = Channel.chatbot.text_components.where(:topic => @topic)
+      @text_component = Text::Sorter.sort(text_components, {}).first
 
       if @text_component
         @next_question_answer = @text_component.question_answers.first
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -1,12 +1,16 @@
 class ChatfuelController < ApplicationController
   def show
-    chatbot_channel = Channel.chatbot
+    @topic = Topic.find_by(name: params[:topic])
 
-    @text_component = Text::Sorter.sort(chatbot_channel.text_components, {}).first
+    if @topic
+      @text_component = Text::Sorter.sort(@topic.text_components, {}).first
 
-    if @text_component
-      @next_question_answer = @text_component.question_answers.first
-      render json: json_response
+      if @text_component
+        @next_question_answer = @text_component.question_answers.first
+        render json: json_response
+      else
+        render json: {}, status: 404
+      end
     else
       render json: {}, status: 404
     end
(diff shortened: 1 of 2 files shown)


app/controllers/chatfuel_controller.rb
```diff
@@ -1,12 +1,16 @@
 class ChatfuelController < ApplicationController
   def show
-    chatbot_channel = Channel.chatbot
+    @topic = Topic.find_by(name: params[:topic])
 
-    @text_component = Text::Sorter.sort(chatbot_channel.text_components, {}).first
+    if @topic
+      @text_component = Text::Sorter.sort(@topic.text_components, {}).first
```
roschaefer: What about text components for the topic but which are not assigned to the chatbot channel? It feels horribly over-engineered for me, but configuring every text component explicitly to go into the chatbot was required by our editorial team, @drjakob is that correct?

spec/requests/chatfuel_spec.rb
```diff
@@ -8,11 +8,15 @@
...
-      let(:topic_id) { 1 }
-      let(:url) { "/chatfuel/#{topic_id}" }
+      let(:topic_name) { "milk_quality" }
+      let(:url) { "/chatfuel/#{topic_name}" }
       let(:chatbot_channel)   { Channel.chatbot }
       let(:report)            { Report.current }
 
+      describe 'topic doesn\'t exist' do
+        it { is_expected.to have_http_status(404) }
+      end
```
roschaefer: Here should be tests for text components attached to different topics

app/controllers/chatfuel_controller.rb
matthib: You are right, I did not take this into consideration, but only the text_components for the current topic which are assigned to the chatbot channel should be displayed.

spec/requests/chatfuel_spec.rb
matthib: @roschaefer  I more tests in the newest commit.

spec/requests/chatfuel_spec.rb
```diff
@@ -8,16 +8,54 @@
...
+            channels: [chatbot_channel],
+            main_part: "The main part",
+            id: 1
+          )
+        end
+
+        it { is_expected.to have_http_status(404) }
+      end
+
+      describe 'text compoent with wrong channel' do
```
roschaefer: typo (x2)

spec/requests/chatfuel_spec.rb
```diff
@@ -8,16 +8,54 @@
...
+            report: report,
+            topic: topic,
+            channels: [Channel.sensorstory],
+            main_part: "The main part",
+            id: 1
+          )
+        end
+
+        it { is_expected.to have_http_status(404) }
+      end
```
roschaefer: Please add a test in wich you
1. Create text component for topic `adorable_kitten`
2. Call the controller for topic `adorable_kitten`
3. Receive the expected text component

matthib: I added the test in the latest commit, is this what you requested?

roschaefer: exactly :smile_cat:

PR #403, opened 2017-06-12
335 390 different reports

### LINKED ISSUE #335: Chatfuel: distinguish between reports SOLVED?
As a reporter
I want the storyboard API to have an identifier for different reports
in order to send chatfuelinformations about one specific report.

### LINKED ISSUE #390: Generate a Sensorstory for every Report
As a reporter
I want to have a sensorstory generated for every of my reports
in order to write different Sensorstories.

roschaefer:
close #335
close #390

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -14,16 +14,13 @@ class Report < ActiveRecord::Base
     Report.first
   end
 
-  def active_sensor_story_components(opts={})
-    Channel.sensorstory.text_components.select {|c| c.active?(opts) }
-  end
-
   def visible_sensor_story_components(opts={})
     active_sensor_story_components(opts).select {|c| c.published? }
   end
 
-  def active_text_components(opts={})
-    text_components.select {|c| c.active?(opts) }
+  def active_sensor_story_components(opts={})
+    result = text_components.includes(:channels)
+    text_components.select {|c| c.channels.include?(Channel.sensorstory) && c.active?(opts) }
   end
 
   def archive!(intention: :real)
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -18,9 +18,12 @@ class Report < ActiveRecord::Base
     active_sensor_story_components(opts).select {|c| c.published? }
   end
 
+  def active_chatbot_components(opts={})
+    active_components(opts).select {|c| c.channels.include?(Channel.chatbot) }
+  end
+
   def active_sensor_story_components(opts={})
-    result = text_components.includes(:channels)
-    text_components.select {|c| c.channels.include?(Channel.sensorstory) && c.active?(opts) }
+    active_components(opts).select {|c| c.channels.include?(Channel.sensorstory) }
   end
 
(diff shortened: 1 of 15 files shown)


app/models/report.rb
```diff
@@ -46,4 +46,11 @@ def end_date
       start_date + duration.days
     end
   end
+
+  private
+
+  def active_components(opts={})
+    result = text_components.includes(:channels)
+    text_components.select {|c| c.active?(opts) }
+  end
```
tillprochaska: This isn’t directly related to the issues solved in this PR, but to #311. If I’m right, sensorstories as well as the chatbot API should both only include published text components. So it might be a good idea to filter for published text components in the `active_components` method in order not repeat ourselfs. The `visible_sensor_story_components` would then be superfluous, but the code would be way cleaner imo. (Yep, I know I have introduced that new `visible_*` method…)

If there are any plans for the future (?!) for a preview features that includes unpublished text components as well, it might still be a good idea to have a single method, that optionally returns unpublished components as well.

In any case: The chatfuel API should return published components only as well, right? Would be great to hear your thoughts on this, I’ll reopen #311 then and work this out.

app/views/text_components/_form.html.haml
```diff
@@ -3,14 +3,11 @@
 
   .form-inputs
 
-    = f.association :report, selected: @report.id
-
```
tillprochaska: Are you sure we should remove that field? I think @drjakob and his team are depending on it to move text components between reports…

tillprochaska: Oh, I’m sorry, that select field does still exists…

tillprochaska:
Just a general note: I think the navigation ux is unclear now. There a elements that are not grouped under a single report (events, actuators, chains), sensors, text components and triggers belong to a single report, though. This isn’t clear from the navigation.

And while users will stay in the respective report namespace when navigating to the index page for text components, triggers or sensors, the current report will be reset when navigating to the chains, actuators or events index, although the users hasn’t explicitly switched between reports.

app/models/report.rb
roschaefer: good point. Is there any situation where we want bot `active` and `unpublished` text components? I'm going to change this..

roschaefer: This is a good example, why PR reviews are a good thing to do

roschaefer:
It's correct, the report will be lost, when the user navigates to anything that is not nested under `reports/:report_id`. I guess, we're going in the direction to nest everything. I mentioned that in the last PR but I think it would be good to open another ticket for that.

app/models/report.rb
tillprochaska: I think generally we will only need active components. However, it’s definitely  a good idea to have an option to include unpublished components as well, in case we need to preview them before publishing in the future. Should default to published text components only, though.

In my opinion, the benefits of a more generally applicable method justify the work that has to be done in order to accomplish this, even if that feature isn’t currently on our list.

PR #425, opened 2017-06-17
407 smaxtec temperature sensor value [WIP]

### LINKED ISSUE #407: Get Smaxtec Temperature as a Sensor Value
As a reporter
I want the system to get the temperature of a cow from the inside sensor (call the Smaxtec API) and generate sensor values based on this readings on a regular basis (cronjob) 
in order to better report about cows.

matthib:
Get temperature sensor value from Smaxtec API and create Sensor Readings in the system for it (first version).

This still has issues with the seed data & the test.

--

Close #407


app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,46 @@ class SmaxtecApi
   SMAXTEC_API_PASSWORD = Rails.application.secrets.smaxtec_api_password
   SMAXTEC_API_BASE_URL = 'https://api-staging.smaxtec.com/api/v1'
 
-  def get_temperature
+  def get_sensor_readings
+    #define animal_id (smaxtec api identifier) for cows: topic_id => animal_id
+    animal_id = { 1 => '5722099ea80a5f54c631513d' }
+
+    # Assuming 1 report = 1 cow, otherwise get just the specified reports
+    Report.all.each do |report|
```
roschaefer: use `find_each`

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,46 @@ class SmaxtecApi
...
+      if animal_id[report.id]
+        sensors = Sensor.where(report_id: report.id, smaxtec_sensor: true)
```
roschaefer: You can combine these two lines:
```ruby
Sensor.where( ... bla ..).each do |sensor|
```

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,46 @@ class SmaxtecApi
...
+        if sensors
+          sensors.each do |sensor|
+            # temperature sensor
+            if sensor.sensor_type_id == 1
```
roschaefer: no magic numbers!

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,46 @@ class SmaxtecApi
...
+              # get temperature data from smaxtec api
+              temperature = get_temperature(animal_id[report.id])
```
roschaefer: I don't understand `animal_id[report.id]`. Apparently this is a hash and you want to get `nil` if it's not `1`? What do you try to accomplish?

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,46 @@ class SmaxtecApi
...
+
+              if temperature
+                # create new sensor readig for sensor
+                # compare temperature with latest sensor reading and only create new sensor reading if value has changed?
+                sensor_reading = Sensor::Reading.new(sensor_id: sensor.id, calibrated_value: temperature, uncalibrated_value: temperature)
```
roschaefer: Is there no unique `id` in the smaxtec database? You can use that to avoid duplicates.

db/static_seeds.rb
```diff
@@ -8,3 +8,6 @@
   {start_date: "2017-05-28", name: "Bio", video: nil, duration: nil},
   {start_date: "2017-05-28", name: "Konventionell", video: nil, duration: nil}
 ])
+Sensor.create!([
```
roschaefer: Does this really have to go into the `static_seeds`? Only include those records, that you want to have in all tests.

lib/tasks/smaxtec_api.rake
```diff
@@ -4,4 +4,10 @@ namespace :smaxtec_api do
     smaxtec_api_controller = SmaxtecApi.new
     puts smaxtec_api_controller.get_temperature
   end
+
+  desc "get sensor readings"
+  task :get_sensor_readings => :environment do
```
roschaefer: Usually I try to avoid these all to generic `get_... ` or `set_... ` methods in favour of a more self-explaining name.

spec/lib/smaxtec_api_spec.rb
```diff
@@ -16,7 +16,21 @@
   describe '.get_temperature' do
     it 'should read the current temperature of a specified animal from the Smaxtec API' do
       VCR.use_cassette('smaxtec_api', :record => :once) do
-        expect(smaxtec_api.get_temperature()).to be_kind_of(Float)
+        expect(smaxtec_api.get_temperature('5722099ea80a5f54c631513d')).to be_kind_of(Float)
```
roschaefer: How about this expectation:
```ruby
expect { smaxtec_api.get_temperature(...) }.to change {Sensor::Reading.count}.from(0).to(1)
```

spec/lib/smaxtec_api_spec.rb
```diff
@@ -16,7 +16,21 @@
...
+      end
+    end
+  end
+
+  describe '.get_sensor_readings' do
+    let(:report) { Report.find(1) }
+    let(:sensor_type) { SensorType.find(1) }
```
roschaefer: magic numbers again

spec/lib/smaxtec_api_spec.rb
```diff
@@ -16,7 +16,21 @@
...
+    let(:temperature_sensor) { Sensor.find_by(name: 'Kuh Berta Temperature') }
+
+    it 'should get the sensor readings from the Smaxtec API and create them in the system' do
+      VCR.use_cassette('smaxtec_api', :record => :once) do
+        number_sensor_readings = temperature_sensor.sensor_readings.count
+        smaxtec_api.get_sensor_readings
```
roschaefer: same comment like above

roschaefer:
Usually, the reviewer would merge the PR when approved and all tests pass. If you want to signal that this PR should not be merged and you just want early feedback, you can prepend `[WIP]` to the name of the PR.

app/lib/smaxtec_api.rb
matthib: Would sensor.sensor_type.property == 'Temperature' be better?

app/lib/smaxtec_api.rb
matthib: I assume that a report is associated with a cow (e.g. 'Report Kuh Berta'). In the Smaxtec API the cow is identified by an animal_id. So at the beginning I define the animal_id hash array(?) and when iterating over every report I want to query the Smaxtec API with the associated animal_id to the report. But I only do this if the report has actually an animal_id (cow) associated with it (see line 14), because I don't know if its always the case.

app/lib/smaxtec_api.rb
matthib: You mean unique Id for each sensor reading to avoid multiple entries for one sensor reading? I don't think so, but I will have to look at the API again.

db/static_seeds.rb
matthib: Maybe it's wrong to put it in the static_seeds. My intention is that I need to have all the sensors for all the cows present (created before) when I query the Smaxtec API and write the Sensor Readings for the associated Sensor.
Is the static_seeds file only the place for the seeds you expect in the test?

roschaefer: Well, you can't assume the sensors to be present. If a sensor is missing that's a good test case.

app/lib/smaxtec_api.rb
roschaefer: If they use a relational db, they have a primary identifier. They should return that, it would be strange if not.

app/lib/smaxtec_api.rb
roschaefer: How about querying over all sensors that have a smaxtec animal_id?

matthib: So instead of giving the sensors a bool smactec_api flag, give them an animal_id and iterate over all the sensors with a present animal_id? That would work as well :+1:

db/static_seeds.rb
matthib: Yes, but the sensors need to be created somewhere and I thought the static_seeds file was the right place for that. If each sensor gets an animal_id (see above) I need to create the specific sensor with all the parameters at some point (preferably when setting up the whole system, before the querying of the API and creating of sensor readings starts). Where else can I create the sensors?

matthib: I didn't think about it first, but the sensors can also be created in the backend by a reporter. The new sensor form would then need a field for the animal_id and a validation that there can only be one animal_id together with a specific sensor type (e.g. only one temperature sensor for a specific animal).

roschaefer: You create sensors in tests just as usual, ie. `create(:sensor, smaxtecs_sensor: true)`, for production, the best option would be to have a user interface where you can create sensors (and provide the `animal_id`)

drjakob:
Yes, user interface would be great.

-- 
Dr. Jakob J. E. Vicari
freier Wissenschaftsredakteur
post@jakobvicari.de
+49 (0)4131 720 50 27

Wissenschaftsjournalist des Jahres 2015, Medium Magazin
Sieger Formatfestival 2015, Medieninnovationszentrum Babelsberg
Nominiert Nannen-Preis 2016, Web-Reportage

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,32 @@ class SmaxtecApi
   SMAXTEC_API_PASSWORD = Rails.application.secrets.smaxtec_api_password
   SMAXTEC_API_BASE_URL = 'https://api-staging.smaxtec.com/api/v1'
 
-  def get_temperature
+  def update_sensor_readings
+    Sensor.where.not(animal_id: [nil, ""]).each do |sensor|
+      # temperature sensor
+      if sensor.sensor_type.property == 'Temperature'
```
roschaefer: This line hard-wires the code with certain records in the database. Ie. you expect data to be present in the database. This is sth. you should try to avoid. In this case, it's inevitable. But you can make the best out of it:
1. Refactor
```ruby
class Sensor::Type
  enum property: [:temperature, ...]
end

class Sensor
  delegate :temperature?, to: :sensor_type
end
```
2. Run
```
bin/rails g migration AddSensorTypes
```
3. Create the sensor types in the database (if they don't exist already)
4. Write one rspec test like:
```ruby
it 'sensor type temperature is in the database' do
  expect(Sensor::Type.temperature.count).to be > 1
end

File: app/lib/smaxtec_api.rb [hunk 12]
```diff
@@ -5,21 +5,32 @@ class SmaxtecApi
...
-      unit = 'degree_celsius'
       temp_data = send_api_request('/data/query', { :animal_id => animal_id, :metric => metric, :from_date => Time.now.to_i - 3600, :to_date => Time.now.to_i, :aggregation => 'hourly.mean' })
-
       if temp_data && temp_data['data'].count() > 1
-        # TODO: Implement Sesor readings instead of rendering Temperature via JSON
-        #sensor_reading = Sensor::Reading.new(sensor_id: 4, calibrated_value: temp_data['data'].last[1], uncalibrated_value: temp_data['data'].last[1])
-        #sensor_reading.save
         return temp_data['data'].last[1]
       else
         return false
```
roschaefer: return a `Sensor::Reading` instead

roschaefer: or `nil`

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,32 @@ class SmaxtecApi
...
     jwt_request = get_jwt
 
     if jwt_request
       @jwt = JSON.parse(jwt_request)['token']
-      organisation_id ='5721e3f8a80a5f54c6315131';
-      animal_id = '5722099ea80a5f54c631513d' # name = Arabella
+      #animal_id = '5722099ea80a5f54c631513d' # name = Arabella
       metric = 'temp'
-      unit = 'degree_celsius'
       temp_data = send_api_request('/data/query', { :animal_id => animal_id, :metric => metric, :from_date => Time.now.to_i - 3600, :to_date => Time.now.to_i, :aggregation => 'hourly.mean' })
```
roschaefer: Make sure, the new sensor reading will get some attribute that is unique both in our database and in the Smaxtec database

app/lib/smaxtec_api.rb
```diff
@@ -5,21 +5,32 @@ class SmaxtecApi
...
+    Sensor.where.not(animal_id: [nil, ""]).each do |sensor|
+      # temperature sensor
+      if sensor.sensor_type.property == 'Temperature'
+        # get temperature data from smaxtec api
+        temperature = get_temperature(sensor.animal_id)
+
+        if temperature
+          # create new sensor readig for sensor
+          sensor_reading = Sensor::Reading.new(sensor_id: sensor.id, calibrated_value: temperature, uncalibrated_value: temperature)
+          sensor_reading.save!
```
roschaefer: If you try to `save` a sensor reading that is already there (ie. some kind of unique attribute is there already) the return value of `save` will be false. In that case maybe you can print out the `sensor_reading.errors` and simple continue on. The procedure will stop this script from creating duplicate data

app/views/sensors/show.html.haml
```diff
@@ -8,6 +8,12 @@
 %p
   %strong Address:
   = @sensor.address
+
+- if user_signed_in? && @sensor.animal_id?
```
roschaefer: Is the `animal_id` classified or why is it not displayed unless signed in? Just curious.

matthib: Yeah, I thought it was better to show it only to authenticated users, because the animal_id is not relevant for a standard user.

PR #436, opened 2017-06-19
338 text component table

### LINKED ISSUE #338: TextComponents List: Enhance structure and styling
As a reporter
I want to have a better colored user interface like in UX-Design-Sketch " 161007_SensorJournalismus1-01.png"
in order to have a better overview using the system.

tillprochaska:
bildschirmfoto 2017-06-19 um 20 43 10

close #338


tillprochaska:
There are a few things on my list for this one

- [x] **Status Indicator**: The circle before the headline (draft: blue, fact_checked: orange, published: green), shows a tooltip with the the written publication status on hover
- [ ] **Show sensor conditions for triggers**
- [ ] **Topic**: Should this be shown in the table?
- [ ] **Priority**: Should this be shown in the table? Space and the user’s attention is limited.
- [ ] **Sorting**: Are there any preferred criteria for sorting triggers/text components in the table? 
Currently, after having edited a text component or a trigger, the displayed order has changed in some cases (because tc and triggers are simply returned in the order they are stored/returned by the db) which might be confusing for editors.
Triggers could be sorted by name, the text components of each trigger by name or start day.

drjakob:
status indicator: great
sensor conditions: would be great if not to much work. At the moment we have long, speaking  trigger names.
priority: no
sorting: triggers alphabetically, text components inside: chronologically after days (other sorting methods, eg by used sensor or by priority are too complicate to use i think.)
topic: later (we'll have symbols for the frontend). A filter for topic (in case) would be better than displaying it

app/models/text_component.rb
```diff
@@ -59,4 +59,21 @@ def css_status_class
...
+
+  def channel_icons
+    channel_icon_mapping = {
+      'sensorstory' => 'fa-file-text',
+      'chatbot' => 'fa-comment'
+    }
+
+    icons = []
+
+    channels.each do |channel|
```
roschaefer: Array#compact removes `nil` values from an array. You can remove the if-clause

app/controllers/text_components_controller.rb
```diff
@@ -86,11 +90,19 @@ def set_index_data
       @text_components = TextComponent.includes(:triggers, :question_answers, :channels)
 
       filter_text_components
+      
+      @trigger_groups = @text_components.order('from_day').group_by { |t| t.trigger_ids }
+      @trigger_groups = @trigger_groups.map { |trigger_ids, components| [ Trigger.find(trigger_ids), components ] }.to_h
+      @trigger_groups = @trigger_groups.sort do |a,b|
+        if(a.first.first && b.first.first)
+          a.first.first.name.downcase <=> b.first.first.name.downcase
```
tillprochaska: Hahah okay, let me try to explain…

The table of text components (or trigger groups) should be sorted by the name of the trigger. `@trigger_groups` is a has with an array of trigger_ids as the key and a two-element array containing a collection of triggers and the matching text components.

So when sorting `a.first.first` is refering to the first trigger in the list of triggers for that trigger group. Phew, not very elegant – I’ll clean this up a little tomorrow.

app/models/text_component.rb
tillprochaska: Didn’t know that, thanks!

app/controllers/text_components_controller.rb
roschaefer: `trigger_groups.map(&:name).join(' ')`

roschaefer: ` @trigger_groups = @trigger_groups.sort_by {|group| group.map(&:name).join(' ')}`

roschaefer: technically you should also sort the trigger in a trigger group :thinking:

tillprochaska: Using `sort_by` is not possible as text components without a trigger need to be sorted to the end. However, I have cleaned up the code for the controller and the template a bit and followed your other tips. Hope this is now easier to read and understand.

roschaefer:
@tillprochaska tests are not finished, but you can merge this when it's done

roschaefer:
@tillprochaska please don't forget to add your invested hours

PR #442, opened 2017-06-23
394 chatfuel continue block report

### LINKED ISSUE #394: Chatfuel API: add report id to continue block.
As a chatbot author 
I want to use different continue_blocks for different reports
in order to have stringent talks.

So please add report id to calling block name pattern: Continue_[topic]_[report_id]

matthib:
Added Report Id to Continue Block like in the following pattern: continue_[topic_name]_[report_id]

---
Close #394

### DIFF
--- a/app/controllers/chatfuel_controller.rb
+++ b/app/controllers/chatfuel_controller.rb
@@ -1,13 +1,18 @@
 class ChatfuelController < ApplicationController
   def show
-    @topic = Topic.find_by(name: params[:topic])
-    if @topic
-      relevant_text_components = @report.active_chatbot_components.select {|c| c.topic_id == @topic.id}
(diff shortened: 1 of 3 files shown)


app/controllers/chatfuel_controller.rb
```diff
@@ -1,13 +1,18 @@
 class ChatfuelController < ApplicationController
   def show
-    @topic = Topic.find_by(name: params[:topic])
-    if @topic
-      relevant_text_components = @report.active_chatbot_components.select {|c| c.topic_id == @topic.id}
-      @text_component = Text::Sorter.sort(relevant_text_components, {}).first
+    @report = Report.find_by_id(params[:report_id])
```
roschaefer: this is done twice: https://github.com/roschaefer/story.board/blob/master/app/controllers/application_controller.rb#L16

PR #445, opened 2017-06-25
443 prevent duplicate smaxtec sensor readings

### LINKED ISSUE #443: Sensor::Reading: No duplicate Smaxtec sensor readings
Make sure we don't get duplicate readings from Smaxtec.

Either:
1. With a unique constraint of `timestamp` and `sensor` in the sensor readings table
or
2. Checking manually in the rake task for a duplicate smaxtec api value (e.g. via timestamp)

matthib:
Prevent duplicate smaxtec sensor readings by adding smaxtec_timestamp to sensor reading table & check if combination of smactec_timestamp and sensor_id already exists before creating new sensor reading.

---
Close #443

### DIFF
--- a/app/lib/smaxtec_api.rb
+++ b/app/lib/smaxtec_api.rb
@@ -28,11 +28,16 @@ class SmaxtecApi
     @jwt = JSON.parse(jwt_request)['token']
     #animal_id = '5722099ea80a5f54c631513d' # name = Arabella
     temp_data = send_api_request('/data/query', { :animal_id => sensor.animal_id, :metric => metric, :from_date => Time.now.to_i - 3600, :to_date => Time.now.to_i })
+
     if temp_data && temp_data['data'].count > 1
       last_entry = temp_data['data'].last
       timestamp = last_entry[0]
-      value = last_entry[1] # WTF?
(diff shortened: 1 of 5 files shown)


app/lib/smaxtec_api.rb
```diff
@@ -28,9 +28,16 @@ def last_smaxtec_sensor_reading(sensor)
...
     #animal_id = '5722099ea80a5f54c631513d' # name = Arabella
     temp_data = send_api_request('/data/query', { :animal_id => sensor.animal_id, :metric => metric, :from_date => Time.now.to_i - 3600, :to_date => Time.now.to_i })
+
     if temp_data && temp_data['data'].count > 1
-      value = temp_data['data'].last[1] # WTF?
-      return Sensor::Reading.new(sensor: sensor, calibrated_value: value, uncalibrated_value: value)
+      last_entry = temp_data['data'].last
+      timestamp = last_entry[0]
+      value = last_entry[1]
+      if !Sensor::Reading.find_by(sensor_id: sensor.id, smaxtec_timestamp: timestamp)
```
roschaefer:
```ruby
return Sensor::Reading.find_or_create_by(sensor_id: sensor.id, smaxtec_timestamp: timestamp) do |reading|
  reading.calibrated_value = value
  reading.uncalibrated_value = value
end
```

PR #455, opened 2017-07-07
added ph value and movement to sensor types

### LINKED ISSUE #392: Read Smax pH-Value
As a reporter
I want to read ph-value from the inside
in order to better report about cows.

### LINKED ISSUE #393: Read Smax Movement
As a reporter
I want to read Movement (1-100)  from the inside
in order to better report about cows.

matthib:
This commit adds the sensor types for ph Value und and movement together with their Smaxtec API equivalents.

I was not 100% about the Smaxtec value for movement, but I think it is act which stands for activity probably. There is also the value act_index which seems to also be an activity index, but act_index is measured in percent and act is measured metrically.
@drjakob Should I double check which is the right value with the Smaxtec guys?

Also when digging through the data of the different cows I found out that not every cow has the same metrics / data available at their sensors. Some don't have ph value and some don't have a movement value available, so it might happen that we don't get all the metrics from all the cows.

---
Close #393 
Close #392


drjakob:
Te ph sensors are just mire expensive, so not all of the cows have them. I thing act-index will be right (movement on a 0% to 100% scale), but please check this with smaxtec.

PR #456, opened 2017-07-12
[WIP] 377 highlight markup

### LINKED ISSUE #377: Create Text Components: Show Sensor Name in Blue Box instead of real markup
As reporter
I want to see a blue box with sensorname in tc-edit instead of markup with sensor-id
in order to better understand markup and avoid mistakes.

tillprochaska:
bildschirmfoto 2017-07-12 um 17 08 03

close #377


roschaefer:
why?

tillprochaska:
This solution was way too unstable. Fixed the issue using a different approach in #502.

PR #466, opened 2017-07-17
391 read smaxtec events

### LINKED ISSUE #391: Read SmaxEvents
As a reporter
I want to read alerts from Smaxtex-API to one sensor

EVENT: 101: "drink cycles" –> sensor drink cycles
EVENT 102: "drink cycle increase", –> sensor drink cycle increase
EVENT 104: "temperature increase" –> SENSOR temperature increase,
etc...
in order to use the analytics of this enterprise for my project.

matthib:
This pull requests reads smaxtec events for specific cows (at the moment the events from the past 30 days because there are not many events) and creates sensor readings (value = 1) for the specific sensor type (1 event = 1 sensor type at the moment) at timestamp at which the event occurred.

---

Close #391


drjakob:
Great. Thanks for solving this. But: Is drinking cycle boolean (set to 1) or is this the specific number of drinking cycles per day? Then it should get the value (1 to 15).

app/lib/smaxtec_api.rb
```diff
@@ -31,6 +56,11 @@ def update_sensor_readings
 
 
   def smaxtec_sensor_readings(sensor)
```
roschaefer: I consider the mixture of `event` and `sensor_reading` code smell.

Create two separate methods for events and sensor readings, please.

matthib:
@drjakob I don't know about the drinking cycle value, I will have to ask Smaxtec about that!

app/lib/smaxtec_api.rb
matthib: Yeah, I wasn't sure about this. I didn't want too much repition, so I mixed the two use cases, but I guess you are right: they should be separated into own methods.

drjakob:
Please don't forget to add the new sensor types types in our list in the front end!

matthib:
@drjakob To create the sensor types I already have the rake task, which creates all new sensor types in the database, so they can be shown in the frontend when creating / editing a sensor.

app/lib/smaxtec_api.rb
```diff
@@ -32,8 +106,11 @@ def update_sensor_readings
...
     metric = PROPERTY_MAPPING[sensor.property]
+    event = EVENT_MAPPING[sensor.property]
     unless metric
-      puts "Sensor #{sensor.name} has unrecognized property: #{sensor.property}"
+      if !event
```
roschaefer: use unless

but apart from that, why do we return `nil`, if metric is nil but event is not nil?

matthib: Because in this method we only want to fetch the sensor readings and not the events. Because we iterate over all the sensors (including the event sensors), we need to make sure we don't fetch any event sensor, because the events are handled in the smaxtec_events method.
For debugging reasons (in the log), we want to know if there is a sensor which has an unrecognized property, this is why we iterate over all available sensors. But we need to make sure we don't show the error message, if this is an event sensor, because in this case it isn't an unrecognized property, but just an event which don't care about in this method.

PR #469, opened 2017-07-18
Feature: JSON-endpoint for sensor readings for Front-End

### LINKED ISSUE #515: JSON-API: Sensor readings
Optimization of JSON API for DataApp

DavidLichtenberger:
close #515


features/sensor_readings/sensor_readings_json_frontend.feature
```diff
@@ -0,0 +1,100 @@
...
+      | 24437 | 2017-07-15 08:39 | 38.353           | 38.353             |
+      | 24419 | 2017-07-15 07:49 | 39.459           | 39.459             |
+      | 24413 | 2017-07-15 07:39 | 39.519           | 39.519             |
+      | 24408 | 2017-07-15 07:29 | 39.489           | 39.489             |
+      | 24407 | 2017-07-15 07:19 | 39.489           | 39.489             |
+
+    And I send and accept JSON
+
+  Scenario: Prepare sensor data (use average for 'throttle', cut off at 'since') and send as json
+    When I send a GET request to "reports/1/sensors/95.json?since=2017.07.17%2000:00&throttle=60"
```
roschaefer: How about this route: `reports/1/sensors/95/sensor_readings.json?since=2017.07.17%2000:00&throttle=60` ?

roschaefer:
This is a beautiful cucumber featue. Thank you @DavidLichtenberger !

Only: What is the meaning of `throttle`'? I don't quite get that.

features/sensor_readings/sensor_readings_json_frontend.feature
```diff
@@ -0,0 +1,96 @@
...
+      | 24443 | 2017-07-15 08:49 | 38.561           | 38.561             |
+      | 24437 | 2017-07-15 08:39 | 38.353           | 38.353             |
+      | 24419 | 2017-07-15 07:49 | 39.459           | 39.459             |
+      | 24413 | 2017-07-15 07:39 | 39.519           | 39.519             |
+      | 24408 | 2017-07-15 07:29 | 39.489           | 39.489             |
+      | 24407 | 2017-07-15 07:19 | 39.489           | 39.489             |
+    And I send and accept JSON
+
+  Scenario: Prepare sensor data (use average for 'throttle', cut off at 'since') and send as json
+    When I send a GET request to "reports/1/sensors/95.json?since=2017.07.17%2000:00&throttle=60"
```
roschaefer: how about:
```
reports/1/sensors/95/sensor_readings.json?since=2017.07.17%2000:00&throttle=60
```

DavidLichtenberger: The alternative URL for the endpoint works fine for me, thanks.

As to throttling: To reduce the performance cost of drawing the curves, I am suggesting to reduce the total amount of data (if necessary). I was thinking it might work like this: 
The total entries of the database get reduced to one entry per specified interval (in the feature it's 60 minutes, but it could be anything) by taking all entries within that interval and outputting their average as a new (virtual?) entry. 
For example:
2017-17-16 00:10 | 2.0
2017-17-16 00:30 | 3.0
2017-17-16 00:45 | 1.0
would become
2017-17-16 00:00 | 2.0

If there is no data in the window, no new entry gets created.

... to be entirely correct we would have to window the average around the new entry (`newEntry.time - 0.5 *interval` to `newEntryTime + 0.5 * interval`), but I don't think we need to be this precise.

I could just as well do the transformation after getting the JSON, while processing it for the front-end, if you think that's better ... I just thought it might improve performance (if only by transmitting less JSON) on the client side.

drjakob:
Great that the collaboration seems to work!

features/sensor_readings/sensor_readings_json_frontend.feature
roschaefer: If `throttle` is just a performance optimization, I would suggest to follow the [KISS design principle ](https://en.wikipedia.org/wiki/KISS_principle) and avoid [premature optimization](http://wiki.c2.com/?PrematureOptimization).

roschaefer: How about just leaving it out for later?

DavidLichtenberger:
Ok, give with me. Let's wait and see if it kills the `<canvas>`.

drjakob:
@roschaefer What to do with this one? Merge?

roschaefer:
Don't merge. Either close or wait until #543 closes it. #543 includes the commit from this PR.

drjakob:
@sebroeder and @matthib please review this. URGENT!

PR #471, opened 2017-07-18
close #373, needs manual intervention

### LINKED ISSUE #373: UPDATE Topics-Items
As a reporter (Björn)
I want a Topic List (in German!) as follows
in order to fit the editorial planing.

*TOPICS*
Breaking News
Hofleben
Abkalbung
Gesundheit
Backup-Kuh
Technologie
Kalb
Milchmenge
Milchqualität
Fütterung
Kuhalltag
Sozialverhalten
Brunst_und_Besamung
Politik_Ethik
Wirtschaft (Milchpreis)
Kurioses

roschaefer:
close #373

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -123,7 +123,7 @@ ActiveRecord::Schema.define(version: 20170719205918) do
 
   create_table "sensor_types", force: :cascade do |t|
     t.string   "property"
-    t.string   "unit"
+    t.string   "unit",           default: ""
     t.datetime "created_at",                       null: false
     t.datetime "updated_at",                       null: false
     t.decimal  "min",            default: "-20.0"
@@ -158,8 +158,8 @@ ActiveRecord::Schema.define(version: 20170719205918) do
     t.integer  "to_day"
     t.integer  "report_id"
     t.integer  "topic_id"
-    t.integer  "publication_status", default: 0
     t.integer  "assignee_id"
+    t.integer  "publication_status", default: 0
     t.datetime "created_at"
     t.datetime "updated_at"
     t.index ["assignee_id"], name: "index_text_components_on_assignee_id", using: :btree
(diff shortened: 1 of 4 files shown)


lib/tasks/add_german_topics.rake
```diff
@@ -0,0 +1,43 @@
+namespace :sensorstory do
+
+  desc 'Add or rename a list of german topics'
+  task add_german_topics: :environment do
+    Topic.transaction do
+      not_matched = [
```
roschaefer: @drjakob what happens with these topics?

drjakob: 3 things: 1. Structure content (eg for search), 2. Control Chatbot, 3. Define Icons for Timestamps

roschaefer: no, I mean, there are three topics that are not in the list of #373

roschaefer:
```
    "movement",
    "temperature",
    "noise"
```

drjakob: Just delete them from text components.

roschaefer: Ok, `movement`, `temperature` and `noise` had 2, 3 and 3 text components.

db/schema.rb
```diff
@@ -139,7 +139,6 @@
     t.float    "max_value"
     t.float    "min_value"
     t.datetime "calibrated_at"
-    t.boolean  "smaxtec_sensor"
```
tillprochaska: Is the removal of the `smaxtec_sensor` property intended? Not sure how this is related to issue #373…

drjakob: As far as I know rhe property is given when an animal id is present

roschaefer: @tillprochaska that's a good find @matthib the `schema.rb` was generated. So I believe this is a leftover of the schema.rb that was merge into master? Do we need to care about `smaxtec_sensor`?

PR #478, opened 2017-07-19
452 naming issues intention

### LINKED ISSUE #452: Report: Fix naming issues of enum column `intention`
All the sensory data is divided into currently two sets: One set of data coming from real sensors and one set of generated data for debugging purposes. The data for debugging will be used only for the `preview` version of the report and the actual data for the `present` version of the report.

The naming is very confusing and misleading within the team. Therefore let's fix these naming isssues:

# Renamings
column intention => release
value real => final
value fake => debug
route present => final
route preview => debug

roschaefer:
close #452 

FYI: @DavidLichtenberger


tillprochaska:
@roschaefer: I’ll review this later today!

PR #481, opened 2017-07-20
479 Added range slider for trigger conditions

### LINKED ISSUE #479: BUG Sensor value slider broken
In create tc modal, create with new Trigger the sliders for sensor values are broken.
Should be a slider with two handles to define the range of sensor values.

tillprochaska:
This is a very simple solution with almost no dependencies. It requires browsers to [support the HTML5 range input](http://caniuse.com/#feat=input-range), though.

bildschirmfoto 2017-07-20 um 18 08 44

close #479


roschaefer:
https://github.com/LeaVerou/multirange

PR #487, opened 2017-07-20
417 min max fraction digits sensor type

### LINKED ISSUE #417: SensorType: Add `min`, `max` and `fractionDigits`
As editor I want 
some Sliders for Sensors should have one digit after comma for some sensortypes (I think this is defined in sensortypes) as follows: 
temperature -10 to +45, Format: "10,2 Grad"
ph 0-14, Format "6,4"
THI (Temperature-Humidity-Index): "72", 1-100, no unit
Milchmenge: 0-80, FORMAT: "57,3 Liter"
Bewegung: 0-100%, FORMAT: "50%"
Humidity: 0-100%, Format: "55%"
Preis: 0-200 Cent, Format: 55 cent

roschaefer:
close #417 

NEEDS MANUAL INTERVENTION

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,20 +27,36 @@
 //= require components/editor
 
 $(function() {
-    autosize($('.field textarea'));
-    autosize.update($('.field textarea'));
+	autosize($('.field textarea'));
+	autosize.update($('.field textarea'));
 });
 
 $(function() {
-    if($('[data-choices]').length > 0) {
-        new Choices('[data-choices]', {shouldSort: false});
-    }
+	if($('[data-choices]').length > 0) {
+		new Choices('[data-choices]', {shouldSort: false});
+	}
 });
 
 $(function() {
-    $('.range').range();
+	$('.range').range();
 
-    $('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
-        $(item).find('.range').range();
-    });
+	$('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
+		$(item).find('.range').range();
+	});
+
+	$('.trigger-conditions').on('cocoon:after-insert', function(e, item) {
+		$('select.choose_sensor').on('change', function() {
+			var options = $(this).find('option:selected')[0].dataset
+
+			$('.range').data('range').reinit({
+				min: options.rangeMin,
+				max: options.rangeMax,
+				step: options.rangeStep,
+				valueFormatter: function(value) {
+					return value + ' ' + options.rangeUnit;
+				}
+			});
+
+		});
+	});
 });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -44,7 +44,7 @@ $(function() {
 		$(item).find('.range').range();
 
 		$(item).find('select.choose_sensor').on('change', function(e, item) {
-			var options = $(this).find('option:selected')[0].dataset
+			var options = $(this).find('option:selected').data();
 
 			$('.range').data('range').reinit({
 				min: options.rangeMin,
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,34 +27,47 @@
 //= require components/editor
 
 $(function() {
-	autosize($('.field textarea'));
-	autosize.update($('.field textarea'));
+    autosize($('.field textarea'));
+    autosize.update($('.field textarea'));
 });
 
 $(function() {
-	if($('[data-choices]').length > 0) {
-		new Choices('[data-choices]', {shouldSort: false});
-	}
+    if($('[data-choices]').length > 0) {
+        new Choices('[data-choices]', {shouldSort: false});
+    }
 });
 
 $(function() {
-	$('.range').range();
+    $('.range').each(function() {
+        var $elm = $(this);
 
-	$('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
-		$(item).find('.range').range();
+        $elm.range({
+            valueFormatter: function(value) {
+                return value + ' ' + $elm.data('unit')
+            }
+        });
+    });
 
-		$(item).find('select.choose_sensor').on('change', function(e, item) {
-			var options = $(this).find('option:selected').data();
+    $('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
+        var $item = $(item);
 
-			$('.range').data('range').reinit({
-				min: options.rangeMin,
-				max: options.rangeMax,
-				step: options.rangeStep,
-				valueFormatter: function(value) {
-					return value + ' ' + options.rangeUnit;
-				}
-			});
+        $item.find('.range').range();
+    });
 
-		});
-	});
+    $(document).on('change', 'select.choose_sensor', function(e) {
+        var $select = $(e.target);
+        var $fields = $select.parents('.nested-fields');
+
+        var options = $select.find('option:selected').data();
+
+        $fields.find('.range').data('range').reinit({
+            min: options.rangeMin,
+            max: options.rangeMax,
+            step: options.rangeStep,
+            valueFormatter: function(value) {
+                return value + ' ' + options.rangeUnit;
+            }
+        });
+
+    });
 });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -42,10 +42,8 @@ $(function() {
 
 	$('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
 		$(item).find('.range').range();
-	});
 
-	$('.trigger-conditions').on('cocoon:after-insert', function(e, item) {
-		$('select.choose_sensor').on('change', function() {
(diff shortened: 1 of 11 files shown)


roschaefer:
@drjakob 
1) The list in #417 has ambiguous or not existing sensor types
2) Some of the current sensor types have a weird `unit` e.g. "Lautstärke"
3) Some of the sensor types are duplicates in English and German

After running the rake task, the resulting sensor types are:
```
Air quality: 
Atmospheric Pressure: Bar
Brightness: 0-10
Button: On/Off
Favs: 
Gas: Alcohol: %
Gas: Methan: %
Gas: Ozone: %
Geophone: 
Gesundheitsstatus: 0 -1 gesund/krank
Humidity: %
Lautstärke: 0-10 - still, leise, mittel, laut, sehr laut
Milchmenge:  Liter
Motion: 0-10
Movement: %
Pedometer: km
pH Value: 
Preis:  Cent
Qualität: Note 1-6
Relative Humidity: %
Temperature: °C
Vibration: 0-10
Volume: 0-10
Zeit: hours
```

lib/tasks/update_sensor_types.rake
```diff
@@ -0,0 +1,21 @@
...
+      SensorType.where(unit: '?...?').update_all(unit: nil)
+      SensorType.where(unit: '1-10').update_all(unit: nil)
+      SensorType.where(unit: '1-100').update_all(unit: nil)
+      SensorType.where(unit: '0-unlimited').update_all(unit: nil)
+      SensorType.where(unit: '').update_all(unit: nil)
+      SensorType.find_by!(property: 'Temperature').update_attributes!(min: -10.0, max: 45.0, fractionDigits: 1, unit: '°C')
+      SensorType.find_by!(property: 'pH Value').update_attributes!(min: -10.0, max: 45.0, fractionDigits: 1, unit: '')
+      SensorType.find_by!(property: 'pH Value').update_attributes!(min: 0.0, max: 14.0, fractionDigits: 1, unit: '')
+
+      #SensorType.find_by('Temperature-Humidity-Index').update_attributes!(min: 1, max: 100, fractionDigits: 0)
```
roschaefer: @drjakob there is no sensor type `Temperature-Humidity-Index`

lib/tasks/update_sensor_types.rake
```diff
@@ -0,0 +1,21 @@
...
+      SensorType.find_by!(property: 'Milchmenge').update_attributes!(min: 0.0, max: 80.0, fractionDigits: 1, unit: ' Liter')
+      SensorType.find_by!(property: 'Movement').update_attributes!(min: 0, max: 100, fractionDigits: 0, unit: '%') # or should it be "Pedometer"?
```
roschaefer: are `Movement' and 'Humidity' the right sensor types? See the comments for alternatives

roschaefer:
@tillprochaska is there any way how I could test this?

app/assets/javascripts/application.js
```diff
@@ -27,20 +27,36 @@
...
+				max: options.rangeMax,
+				step: options.rangeStep,
+				valueFormatter: function(value) {
+					return value +  options.rangeUnit;
+				}
+			});
+
+		});
+	});
+});
```
tillprochaska: You can merge those two event handlers. It doesn’t matter wether you’re setting the event handlers on the select before or after adding it to the DOM.

app/assets/javascripts/application.js
```diff
@@ -27,20 +27,36 @@
...
-});
\ No newline at end of file
+	$('.range').range();
+
+	$('.trigger-conditions').on('cocoon:before-insert', function(e, item) {
+		$(item).find('.range').range();
+	});
+
+	$('.trigger-conditions').on('cocoon:after-insert', function(e, item) {
+		$('select.choose_sensor').on('change', function() {
```
tillprochaska: `$(item).find('.choose_sensor')` will return only the select field in the item to be added. Otherwise you’re matching all sensor select in the document (which is probably not that severe in most cases, but simply not necessary).

lib/tasks/update_sensor_types.rake
```diff
@@ -0,0 +1,26 @@
...
+      SensorType.where(unit: '0-unlimited').update_all(unit: '')
+      SensorType.where(unit: '').update_all(unit: '')
+      SensorType.find_by!(property: 'Temperature').update_attributes!(min: -10.0, max: 45.0, fractionDigits: 1, unit: '°C')
+      SensorType.find_by!(property: 'pH Value').update_attributes!(min: -10.0, max: 45.0, fractionDigits: 1, unit: '')
+      SensorType.find_by!(property: 'pH Value').update_attributes!(min: 0.0, max: 14.0, fractionDigits: 1, unit: '')
+
+      #SensorType.find_by('Temperature-Humidity-Index').update_attributes!(min: 1, max: 100, fractionDigits: 0)
+      SensorType.find_by!(property: 'Milchmenge').update_attributes!(min: 0.0, max: 80.0, fractionDigits: 1, unit: 'Liter')
+      SensorType.find_by!(property: 'Movement').update_attributes!(min: 0, max: 100, fractionDigits: 0, unit: '%') # or should it be "Pedometer"?
+      SensorType.find_by!(property: 'Humidity').update_attributes!(min: 0, max: 100, fractionDigits: 0, unit: '%') # or should it be "Relative Humidity"?
```
roschaefer: @drjakob two ambiguous sensor types - see comments

tillprochaska:
> is there any way how I could test this?

You’re talking about the JavaScript parts, right? The `Range` instance is stored in the `data-range` attribute of the corresponding element, so you can easily access it’s methods, even in the context of a test.

```javascript
$('.range').data('range').value();
$('.range').data('range').value('-10,10');
$('.range').data('range').options;
```

And of course you can test the text/values of the elements added by the range component. For example, in order to test wether the correct unit is shown after you selected a different sensor, you could check for the `.range__info` or `.range__tooltip` elements…

lib/tasks/update_sensor_types.rake
```diff
@@ -0,0 +1,26 @@
...
+      SensorType.find_by!(property: 'Preis').update_attributes!(min: 0, max: 200, fractionDigits: 0, unit: 'Cent')
+
+      SensorType.find_by!(unit: '0 -1 gesund/krank').update_attributes!(min: 0, max: 1, fractionDigits: 0, unit: 'gesund/krank')
+      SensorType.find_by!(unit: '0-10 - still, leise, mittel, laut, sehr laut').update_attributes!(min: 0, max: 10, fractionDigits: 0, unit: 'Geräuschpegel')
+      SensorType.find_by!(unit: 'Note 1-6').update_attributes!(min: 1, max: 6, fractionDigits: 0, unit: 'Note')
```
roschaefer: @drjakob here I match the sensor types by `unit` instead of property and change `min`, `max`, 'fractionDigits` and `unit`

app/assets/javascripts/application.js
roschaefer: How can I "finish" this review?

tillprochaska:
One more thing: Right now, the slider will only be updated with the sensor’s min/max values and unit when the user changes the sensor type. Passing in the sensor defaults if a sensor is selected in the [condition fields view](https://github.com/roschaefer/story.board/blob/417_min_max_fractionDigits_sensor_type/app/views/triggers/_condition_fields.html.haml#L12) should be enough.

roschaefer:
@tillprochaska the user can not change the sensor type but the sensor. I don't quite get what you are saying.

tillprochaska:
@roschaefer: Yeah, I meant the sensor. When a user [edits an existing trigger](http://redaktion.superkuehe.de/reports/1/triggers/97/edit), the condition fields will be rendered server side and not by cocoon. Therefore, the slider will still use options `{min: -20, max: 100, step: 1}`. See [_condition_fields.html.haml](https://github.com/roschaefer/story.board/blob/417_min_max_fractionDigits_sensor_type/app/views/triggers/_condition_fields.html.haml#L12).

roschaefer:
@tillprochaska got it!

roschaefer:
@tillprochaska thank you so much for your support. Make sure you add your invested hours to this PR.

@drjakob this PR kept us busy the whole afternoon and evening. Let's consider the costs when implementing intensive UI for the future.

roschaefer:
@tillprochaska please resolve the review when the build is green

tillprochaska:
@roschaefer: Okay, tests are green now! 🎉

@roschaefer, @drjakob: However, there’s one minor flaw remaining. Some sensors don’t have a real unit (i. e. SI-unit), but kind of description, e. g. (Note 1-6, 0-unlimited).

This breaks the UI (which might still be acceptable), but is probably output in the generated reports as, well, right? Maybe we can completely omit the unit in these cases or have a separate description for those sensors instead?

bildschirmfoto 2017-07-23 um 00 02 58

roschaefer:
@tillprochaska @drjakob I am aware of weird values for units. AFAIK there is no user interface to change sensor types, so sb. manipulated the records directly. The right thing to do is to change the units. `1 Note` is not a valid unit anyway.

PR #488, opened 2017-07-21
Added destroy and reinit methods to Range component

tillprochaska:
@roschaefer: You can now reinitialize the range slider at any time.

```javascript
$('.selector').data('range').reinit({
  min: -10,
  max: -5,
  step: .1
})
```

This will keep the rangeslider’s value or set it to the new min/max value if it doesn’t lie in the value range.

Of course you can also set a new value in the same go:

```javascript
$('.selector').data('range').reinit({
  min: -10,
  max: -5,
  step: .1
}).value('-9,-6')
```

And you can also completely reset the slider to restore the initial state (which is what the `reinit` method uses internally):

```javascript
$('.selector').data('range').destroy()
```

### DIFF
--- a/app/assets/javascripts/components/range.js
+++ b/app/assets/javascripts/components/range.js
@@ -16,27 +16,28 @@ var Range = (function($, multirange) {
         this.$tooltipLow = null;
         this.$tooltipHigh = null;
 
-        this.options = {
+        this.init();
+
+        return this;
+    }
+
+    Range.prototype.init = function(options) {
+        var self = this;
+
+        self.options = {
             min: this.$elm.data('min') || 0,
             max: this.$elm.data('max') || 100,
             step: this.$elm.data('step') || 1,
             value: [this.$min.val() || 0, this.$max.val() || 100].sort(),
         }
 
-        this.init();
-
-        return this;
-    }
-
-    Range.prototype.init = function() {
-        var self = this;
+        // extend default options
+        if(typeof options === 'object') {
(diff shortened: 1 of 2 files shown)


tillprochaska:
@roschaefer: One thing I could imagine to improve the UX as well is to add the corresponding units to the values displayed at the slider.

bildschirmfoto 2017-07-21 um 12 50 31

I’d add an optional configuration option that allows you to pass a function to format values like this.

```javascript
$('.selector').data('range').reinit({
  min: -10,
  max: -5,
  step: .1,
  valueFormatter: function(value) {
    return value + ' °C';
  })
})
```

In the context of the sensor select it could be used like this:
```html
<select class="sensor-select">
  <option data-range-min="0" data-range-max="40" data-range-step="1" data-range-unit="°C" value="id">Sensor Name</option>
  …
</select>
```

```javascript
$('.sensor-select').on('input change', function() {
  
  var config = $(this).find('option:selected').data().range;

  $('.range').data('range').reinit({
    min: config.min,
    max: config.max,
    step: config.step,
    valueFormatter: function(value) {
      return value + ' ' + config.unit;
    })
  });

});
```

roschaefer:
I'm merging this into `417_min_max_fractionDigits_sensor_type` and see how far I get. Looks good so far. Thanks a million

roschaefer:
Looks good so far, an option to pass in a function for the value formatting would be great!

tillprochaska:
@roschaefer: Here you go!

PR #502, opened 2017-07-23
377 highlight markup in the editor

tillprochaska:
highlight 1


app/views/text_components/_form.html.haml
```diff
@@ -66,7 +66,7 @@
             %a.form__section__action--open.btn.btn-sm.btn-secondary{href: '#', 'data-form-toggle': 'open-section'} Edit
 
       .form__section__body
-        .text-editor
+        .text-editor{data: {'editor-data': {sensors: @sensors.map{ |s| [s.id, s.name]}.to_h, events: @events.map{ |e| [e.id, e.name] }.to_h }.to_json }}
```
roschaefer: does this become an attribute with a JSON value?

tillprochaska: Yes, exactly!

tillprochaska:
@roschaefer: You’re probably right with that. Let’s talk about that later during our team conf. This whole jQuery things makes testing a bit difficult.

PR #509, opened 2017-07-24
421 sensory data fallback

### LINKED ISSUE #421: Missing sensory data
As a reporter
I want to define a default text for missing sensory data for every sensor type (instead "missing sensory data --."
In order to better inform my readers in German language.

tillprochaska:
The text renderer now supports optional fallback values for both the sensor and event markup:

```
The value of the sensor is { value(123) | "currently not availbale" }.
```

I also made some minor changes:

- Tooltip position in text editor at cursor position
- The render won’t match markup that has lots of whitespace in it anymore, e. g. `{ value(123) }` and `{ value(123) | "Fallback" }` but not `{  value(123)   }` or `{ value( 123) | "Fallback" }`

@drjakob: The text editor will now ask users for a fallback value when using the sensor and event buttons. Wether this is a good solution for your workflow depends a bit on wether fallback values should be used in most cases or only in a few cases.

@roschaefer: Btw, the the test setup for events is a bit scary… ;)

text_editor_fallback 1

close #421

### DIFF
--- a/app/assets/javascripts/components/editor.js
+++ b/app/assets/javascripts/components/editor.js
@@ -38,9 +38,21 @@ var Editor = (function($, autosize) {
          * @type {function[]}
          */
         this.commands = {
-            'text': function($context) {
-                return $context.data().editorMarkupText;
+            text: function($context) {
+                return $context.data('editorMarkupText');
             },
+            sensor: function($context) {
+                var id = $context.data('editor-markup-sensor');
+                var fallback = prompt('Enter a fallback value that will be output in case that there is no recent sensory data for this sensor.');
+
+                return '{ value(' + id + ')'+ (fallback ? ' | "' + fallback + '"' : '') + ' }';
+            },
+            event: function($context) {
+                var id = $context.data('editor-markup-event');
+                var fallback = prompt('Enter a fallback value that will be output in case that there is no recent data for this event.');
+
+                return '{ date(' + id + ')'+ (fallback ? ' | "' + fallback  + '"' : '') + ' }';
+            }
         };
     };
 
@@ -178,7 +190,7 @@ var Editor = (function($, autosize) {
      */
     Field.prototype.filters = {
         sensors: {
-            pattern: /{\s*value\(\s*(\d+)\s*\)\s*}/g,
+            pattern: /{\s?value\(\s?(\d+)\s?\)(?:\s?\|\s?"(.*)")?\s?}/g,
             markup: function(markup, id) {
                 var cls = ['markup', 'markup--sensor'];
                 if(!this.editor.data.sensors[id]) cls.push('markup--error');
@@ -193,7 +205,7 @@ var Editor = (function($, autosize) {
             }
         },
         events: {
-            pattern: /{\s*date\(\s*(\d+)\s*\)\s*}/g,
+            pattern: /{\s?date\(\s?(\d+)\s?\)(?:\s?\|\s?"(.*)")?\s?}/g,
             markup: function(markup, id) {
                 var cls = ['markup', 'markup--event'];
                 if(!this.editor.data.events[id]) cls.push('markup--error');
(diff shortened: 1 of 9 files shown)


drjakob:
We could try this. I think Fallback-Definition in Sensor-Def would be better for workflow. (Because fallback for temperature sensor will nearly always be the same.)

app/assets/javascripts/components/editor.js
```diff
@@ -282,7 +296,7 @@ var Editor = (function($, autosize) {
 
         if(tooltip && self.isFocussed()) {
             var text   = self.value();
-            self.$mirror.html(text.substring(0, activeMarkup.start) + '<span class="field__markup--active">' + text.substring(activeMarkup.start, activeMarkup.end) + '</span>' + text.substring(activeMarkup.end, text.length));
+            self.$mirror.html(text.substring(0, selection.start) + '<span class="field__markup--active"></span>' + text.substring(selection.end, text.length));
```
roschaefer: sneaky

app/assets/javascripts/components/editor.js
```diff
@@ -38,9 +38,21 @@ var Editor = (function($, autosize) {
          * @type {function[]}
          */
         this.commands = {
-            'text': function($context) {
-                return $context.data().editorMarkupText;
+            text: function($context) {
```
roschaefer: Why do we actually use a global `$context` variable as parameter?

app/assets/javascripts/components/editor.js
```diff
@@ -38,9 +38,21 @@ var Editor = (function($, autosize) {
...
+                return $context.data('editorMarkupText');
             },
+            sensor: function($context) {
+                var id = $context.data('editor-markup-sensor');
+                var fallback = prompt('Enter a fallback value that will be output in case that there is no recent sensory data for this sensor.');
```
roschaefer: change to `in case there is no sensory data at all for this sensor`

app/assets/javascripts/components/editor.js
```diff
@@ -38,9 +38,21 @@ var Editor = (function($, autosize) {
...
+
+                return '{ value(' + id + ')'+ (fallback ? ' | "' + fallback + '"' : '') + ' }';
+            },
+            event: function($context) {
+                var id = $context.data('editor-markup-event');
+                var fallback = prompt('Enter a fallback value that will be output in case that there is no recent data for this event.');
+
+                return '{ date(' + id + ')'+ (fallback ? ' | "' + fallback  + '"' : '') + ' }';
```
roschaefer: So this will be inserted as soon as sb. enters a sensor? I like the solution. Only concern: Will the extra effort start to annoy me at some point?

app/assets/javascripts/components/editor.js
tillprochaska: The `$context` variable is the element that triggered the text insertion command (see [this line](https://github.com/roschaefer/story.board/blob/f94e157d58260064ebfc587bdcb8e633da24257f/app/assets/javascripts/components/editor.js#L82)) so that we can access e. g. the data attributes.

app/assets/javascripts/components/editor.js
tillprochaska: No, only if you insert markup using the sensor or events buttons. You can still write markup manually.

tillprochaska:
Will do tomorrow.

> From the user perspective: The concern that a user has to understand the behaviour doesn't hold if the fallback prompt is inserted straight away into the text. However I imagine that bothering about a fallback might become annoying at some point.

Yes, you’re right about that. That’s what my question in the initial PR description was about. Next time, I’ll wait for feedback before implementing features… ;)

> The text editor will now ask users for a fallback value when using the sensor and event buttons. Wether this is a good solution for your workflow depends a bit on wether fallback values should be used in most cases or only in a few cases.

app/assets/javascripts/components/editor.js
tillprochaska: No, only if you insert markup using the sensor or events buttons. You can still write markup manually.

app/assets/javascripts/components/editor.js
tillprochaska: The `$context` variable is the element that triggered the text insertion command (see [this line](https://github.com/roschaefer/story.board/blob/f94e157d58260064ebfc587bdcb8e633da24257f/app/assets/javascripts/components/editor.js#L82)) so that we can access e. g. the data attributes.

roschaefer: why is it global?

tillprochaska: It isn’t?

Issue #441, opened 2017-06-22
Character Counter


drjakob:
As a reporter
I want to have a character count for main
in order to better text for frontend.

PR #511, opened 2017-07-24
Larger textarea for main content in tc form

### LINKED ISSUE #491: Adjustable text input
As a reporter (Björn)
I want to have a adjustable input field for sensorstory-Main-text
In order to write snd edit longer texts.

tillprochaska:
@drjakob: Okay, I made the main input larger by default. Maybe that’s already enough. Setting it to 8 lines makes the form unecessary long and in the consequence harder to scan, though. Maybe something like this could help to give a quick visual impression of the text fields while still keeping the form short.

bildschirmfoto 2017-07-24 um 19 51 01

And if your editors are looking for an indicator of text length, a character/word count might be a better idea instead (and is btw pretty cheap to implement).

bildschirmfoto 2017-07-24 um 19 48 26

close #491

### DIFF
--- a/app/assets/stylesheets/components/_field.scss
+++ b/app/assets/stylesheets/components/_field.scss
@@ -18,13 +18,17 @@
 }
 
 .field__input, .field__mirror {
-    min-height: 5.75rem;
+    min-height: 2rem + .75rem + 2 * $line-height-base;
     padding: .75rem;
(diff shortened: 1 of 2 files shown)


drjakob:
Character-Count would be great. (Its already an issue: #441).

PR #520, opened 2017-07-27
[WIP]  Improved input fields for manual sensor inputs

### LINKED ISSUE #473: Improved input fields for manual sensor inputs
As a reporter
I want to enter date and time for my manual sensor readings in two seperate fields in german standard
in order to prevent mistakes. 

DATE: day[1-2].month[1-2].[year4]. Default should be actual date.
TIME: Hour[0-23]:Minute[0-59] . Default should be actual time. Seconds should be by default zero.

Please implement 5 select fields.

matthib:
The purpose of this PR is to improve the manual input of sensor readings by offering select fields to select the date / time of the sensor reading (with current date / time preselected) and refreshing the page when an entry has been made, in order to give a visual feedback of the new record (-> show the user that entering the sensor reading worked).

At the moment the text field for date / time has been replaced by select fields with the preselected current time.

The feature that needs to be implemented yet, is the page reload when entering a new sensor reading. This is probably a bit tricky, because the form is set to remote (Ajax) and the create method for the sensor_readings expects / returns json values (https://github.com/roschaefer/story.board/blob/master/app/controllers/sensor_readings_controller.rb#L3).

---
Close #473

### DIFF
--- a/app/assets/javascripts/sensors.coffee
+++ b/app/assets/javascripts/sensors.coffee
@@ -13,5 +13,13 @@ ready = ->
     $('#unhide_sensor_reading_form').hide()
     $('#sensor_reading_form').show()
 
+  $('#sensor_reading_day, #sensor_reading_month, #sensor_reading_year, #sensor_reading_hour, #sensor_reading_minute, #sensor_reading_second').on "change", ->
+    update_created_at()
+
+
+
+update_created_at =  ->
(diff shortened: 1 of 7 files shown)


roschaefer:
@matthib check the two latest commits. The trick is render some javascript when the sensor was created in the controller.

Add you invested hours.

matthib:
@roschaefer Elegant solution and nice to see how easy rails makes it to render the new record, even with javascript. Also the datetime field is an easier solution than mine. I actually searched for something like this, but didn't find it.

PR #524, opened 2017-07-29
[WIP] 513 bring back default pages for text component

### LINKED ISSUE #513: PERFORMANCE: TextComponents#index page load time way too long
The list of tcs is rendered really slow (or not at all) because of pre-rendering of tc-edit-modal.

roschaefer:
close #513


drjakob:
Please be aware that we don't want to have a new "show" step between list and edit.

roschaefer:
took me a while to fix all the cucumber tests

@tillprochaska let's try to stick as much as possible to standard html. E.g. capybara does not recognize the values of the input fields on the text component edit page as `text`

tillprochaska:
@roschaefer: All those text inputs are just basic inputs generated by simpleform?!

app/views/text_components/_form.html.haml
```diff
@@ -1,133 +1,125 @@
...
-        = f.association :channels, collection: Channel.order('name'), selected: text_component.channels.count > 0 ? text_component.channel_ids : Channel.find_by({name: 'sensorstory'}).id, input_html: {'data-choices': true}
-
-
-  .form-actions
-    = f.button :submit, class: 'btn btn-primary'
+.container
+  = simple_form_for([@report, @text_component]) do |f|
+    = f.error_notification
+    .form
+      .row.form__section
```
tillprochaska: @roschaefer: Why did you add all those row/col wrappers here?

roschaefer: Not necessary. I thought I could fix phantomjs with it. Turns out no.

Try out phantom js and replace the `click_regardless_of_overlapping_element`. You will see that the edit buttons are misaligned in the screenshot. I hoped I could fix it with explicit grid. Or with a smaller screen width (force stacked layout). None of those solutions work and in the end I had to use `.trigger('click')` which is only available in `:poltergeist`. Bad.

roschaefer: in the end I decided to keep the layout - we can remove it, if necessary

roschaefer:
what text inputs?

PR #525, opened 2017-07-30
Allow blank animal ids for sensors

### LINKED ISSUE #500: Bug Creating Sensors: Animal ID should never be required
When Creating new Sensors Animal ID is required, but should never be a required field. (Empty field but  qwarnung message"already be taken"

matthib:
This PR solves the issue #500 "Bug Creating Sensors: Animal ID should never be required". The problem was that the Animal Id is validated to be unique for a Sensor Type together with an Animal Id. If the Animal Id is empty for one Sensor Type, there can't exist another similar Sensor Type with an empty Animal Id. 
This is solved by allowing the Animal Id to be blank (empty).

---

Close #500

### DIFF
--- a/app/models/sensor.rb
+++ b/app/models/sensor.rb
@@ -15,7 +15,7 @@ class Sensor < ActiveRecord::Base
   validates :name, presence: true, uniqueness: true
   validates :address, uniqueness: true
   validates :sensor_type, presence: true
-  validates :animal_id, uniqueness: { scope: :sensor_type }
+  validates :animal_id, uniqueness: { scope: :sensor_type }, :allow_blank => true
 
   def name_and_id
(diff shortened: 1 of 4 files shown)


roschaefer:
@matthib merge it when you are OK with this.

roschaefer:
@matthib you are correct. If you can, try to enforce validation with database constraints. I created some [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) to help with that. Unique constaints with scopes are a bit tricky, but we're lucky because `NULL != NULL` in the database. But the blank string is`'' == ''` as I found out in the tests. So I just turn blanks into `nil`. Done.

roschaefer:
merge it @matthib !

PR #528, opened 2017-07-31
406 event activations

### LINKED ISSUE #406: Events: save start/end of a period to the database
In the current implementation, events have a `last_happened_at` property that is updated every time that particular event is triggered. Events are not directly connected to reports.

Therefore, when an event happens and there are text components triggered by that event, they will become active across all reports.

For example, when a calf is born on the organic farm, the respective text components will be shown in all three reports.

I’m pretty sure that this isn’t the desired behaviour. A possible solution would be to have EventTypes and Events, similar to the relation between Sensors and SensorReadings. Every Event belongs to a report and every EventType can have multiple events. A new event is created everytime that EventType ist triggered.

// cc @drjakob

### LINKED ISSUE #431: On/off toggle switch for list of events
As a reporter
I want an on-off toggle switch for every event in events list (https://www.w3schools.com/howto/howto_css_switch.asp)
in order to easily switch my events on or off.

roschaefer:
going with the term `event_activation`

close #406 
close #431


roschaefer:
@drjakob FYI: the events are still not nested under `/reports/`. Are we going to nest them under `/reports/`?
If yes, how shall the currently existing events be connected in the migration? Simply the `Report.current` (id=1)?

drjakob:
Yes, nesting would be fine. Please migrate existing events with report ID=4 (because they are real)

roschaefer:
good. In order to keep PR small, let's merge this before we nest events under report.

drjakob:
sure

drjakob:
This is really needed and important for the current workflow. @sebroeder could you review this?

roschaefer:
Thanks @sebroeder, very detailed code review. I followed your instructions, I think Github is lacking "request review" again button.

sebroeder:
This looks really good, great work 👍🎉. Could you please run `bin/rake erd` before we merge this, so that the Entity-Relation-Diagram is up to date?

app/models/event/activation.rb
```diff
@@ -0,0 +1,36 @@
+class Event::Activation < ApplicationRecord
+  belongs_to :event, required: true
```
sebroeder: `required: true` is deprecated since Rails 5.0, we can use `optional: false` instead. Since we did not [change the default configuraion](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-belongs-to-required-by-default-option) for our app yet, a deprecation warning might actually be a good thing, though. 

> :required
When set to true, the association will also have its presence validated. This will validate the association itself, not the id. You can use :inverse_of to avoid an extra query during validation. NOTE: required is set to true by default and is deprecated. If you don't want to have association presence validated, use optional: true.

drjakob:
Just manually!

PR #538, opened 2017-08-01
Fix context nesting in `text_component_spec.rb`

sebroeder:
The contexts 'given one trigger' and 'given many triggers' where nested below 'triggers empty' which is a logical contradiction. Put them on the same nesting level as 'triggers empty' instead.

Output of `rspec -f documentation` before the change:

```
TextComponent
  #active?
    triggers empty
      should equal true
      given one trigger
        active trigger
          should equal true
        inactive trigger
          should equal false
      given many triggers
        some are active, some are inactive
          should equal false
        all active
          should equal true
```

And after the change:

```
TextComponent
  #active?
    triggers empty
      should equal true
    given one trigger
      active trigger
        should equal true
      inactive trigger
        should equal false
    given many triggers
      some are active, some are inactive
        should equal false
      all active
        should equal true
```

### DIFF
--- a/spec/models/text_component_spec.rb
+++ b/spec/models/text_component_spec.rb
@@ -48,30 +48,34 @@ RSpec.describe TextComponent, type: :model do
   describe '#active?' do
     let(:text_component) { create(:text_component) }
     subject { text_component.active? }
+
     context 'triggers empty' do
       it { is_expected.to be true }
+    end
 
-      context 'given one trigger' do
-        context 'active trigger' do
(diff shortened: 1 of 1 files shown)


roschaefer:
Good!

Could you please add a file `.rspec` with the following content:
```
format --doc
```

I'm not just pushing a commit because this is sth. interesting for all of us @tillprochaska @matthib: 
* Improve rspec tests by writing readable descriptions.

More information: http://www.betterspecs.org/

PR #541, opened 2017-08-02
Disallow indexing of the CMS backend in `robots.txt`

sebroeder:
The CMS backend content that is not yet behind a login shows up in search engine results.
Disallow all crawlers to index the CMS backend. This is the most restrictive setting possible.
Consider this a workaround and not a full replacement for a password protected login.


roschaefer:
LGTM :+1:

PR #542, opened 2017-08-02
[WIP] 513 bring back default pages for text component

### LINKED ISSUE #513: PERFORMANCE: TextComponents#index page load time way too long

roschaefer:
close #513

@tillprochaska I need your help for debugging the `.editor`. Why are the affected methods never being executed?


tillprochaska:
Hey @roschaefer. Sorry for my late reply, I had completely overlooked this one. The solution to the problem is pretty simple. The `.text-editor__fields` element needs to be a child node of the `.text-editor` which was somehow messed up. Subsequently, the text editor fields weren’t initialized.

In order to add a character count to a field, it’s necessary to add a `.field__count` to the respective `.field`.

PR #543, opened 2017-08-03
515 sensor readings json api

### LINKED ISSUE #515: JSON-API: Sensor readings

roschaefer:
close #515 
close #469


features/sensor_readings/sensor_readings_json_frontend.feature
```diff
@@ -0,0 +1,62 @@
...
+      | 24443 | 2017-07-15 08:49 | 38.561           | 38.561             | final   |
+      | 24437 | 2017-07-15 08:39 | 38.353           | 38.353             | final   |
+      | 24419 | 2017-07-15 07:49 | 39.459           | 39.459             | final   |
+      | 24413 | 2017-07-15 07:39 | 39.519           | 39.519             | final   |
+      | 24408 | 2017-07-15 07:29 | 39.489           | 39.489             | final   |
+      | 24407 | 2017-07-15 07:19 | 39.489           | 39.489             | final   |
+
+  Scenario: Prepare sensor data (between 'since' and 'to') and send as json
+    Given I send and accept JSON
+    When I send a GET request to "reports/1/sensors/95/sensor_readings?from=2017.07.17%2014:40&to=2017.07.18%2012:19"
```
roschaefer: @DavidLichtenberger I renamed the URL param `since` to `from`

features/sensor_readings/sensor_readings_json_frontend.feature
```diff
@@ -0,0 +1,62 @@
...
+    Then the JSON response should be:
+    """
+    [
+    {"id": 24527, "created_at": "2017-07-18T12:19:00.000Z", "calibrated_value": 39.008, "uncalibrated_value": 39.008, "release": "final" },
+    {"id": 24521, "created_at": "2017-07-18T12:09:00.000Z", "calibrated_value": 39.068, "uncalibrated_value": 39.068, "release": "final" },
+    {"id": 24516, "created_at": "2017-07-18T11:59:00.000Z", "calibrated_value": 39.188, "uncalibrated_value": 39.188, "release": "debug" },
```
roschaefer: @DavidLichtenberger 
1. The attribute for the timestamp is called `created_at`
2. I hope you don't mind the extra attributes like `release` (if it's a generated sensor reading for debugging purposes or not)
3. You can also filter for `release`

config/routes.rb
```diff
@@ -32,11 +32,12 @@
       member do
         put :start_calibration
         put :stop_calibration
+
+        resources :sensor_readings, only:[:show, :create, :index], default: { format: :json }
+        post 'sensor_readings/debug', to: 'sensor_readings#debug', default: { format: :json }
```
roschaefer: @kabel42 I moved the `sensor_readings` route up to `/reports/:report_id/sensors/:id/sensor_readings`. If you don't know the correct `:id` of our sensor, just use any number and override it via `address` or `name` in the supplied params.

features/sensor_readings/sensor_readings_json_frontend.feature
DavidLichtenberger: @roschaefer: looks good to me! Is this implemented already? Will try it out as soon as I can (probably tonight) ...

Thanks!

roschaefer: This is not deployed, yet. We have auto-deployment through Travis CI, ie. if this PR is merged into master, it will be online after ~5min. If you subscribe to this thread (which happened because you wrote a comment) you will receive a notification when this PR is merged.

DavidLichtenberger: I see. Thanks again!

roschaefer:
@matthib thanks for looking into this. I should give more context here: You probably get confused about the unrelated changes in the `sensor_reading_controller`. I moved the route for `sensor_readings` into the scope of the sensor: `/reports/:report_id/sensors/:id/sensor_readings` instead of just `/sensor_readings`. I had to fix some cucumber and rspec tests which are either obsolete or point to the wrong url.

The `/sensor_readings` route, apart from being used as an endpoint for the frontend, is used by our custom reporter box. The reporter box comes with certain limits. E.g. arbitrary data *must* be sent in the `data` attribute. That's why we parse the `data` attribute as JSON. Very ugly, probably a security flaw. Anyways, I can't think of another solution if we want to keep the functionality. Another limitation is that the reporter box does not know the sensor `id`. Instead it either sends a `name` of the sensor or the `address`, which is a proprietary address of the physical sensor. The `name` or the `adress` must be entered into the `story.board` by the service team.

It's good to have this knowledge transferred among the team. Probably you now understand the reason for the `address` of a sensor :+1: 

I'll ask @sebroeder. He's going to hate me for the terrible code in `sensor_readings_controller`.

drjakob:
@sebroeder please approve this!

PR #545, opened 2017-08-08
Set the default priority for triggers in the database

sebroeder:
Database default values work great with Rails enums, no custom `after_initialize` callbacks required.

    >> t = Trigger.new
    => #<Trigger id: nil, name: nil, created_at: nil, updated_at: nil, report_id: nil, priority: "medium", validity_period: nil, from_hour: nil, to_hour: nil>


roschaefer:
I thought we covered the situation that a `Trigger` has no priority somewhere in the test suite? Weird. Anyways, I think this is OK for me. Although default `== :medium` is somehow questionable..

PR #552, opened 2017-08-13
Bring back turbolinks

### LINKED ISSUE #514: Performance: Bring back Turbolinks
Turbolinks was removed at some point because of reasons.. find out why and if we can bring it back

roschaefer:
close #514 

~~sledgehammer the cucumber test - I still don't understand why only phantomjs and not chrome fails here and only after we activated turbolinks~~


roschaefer:
:love_letter:

PR #585, opened 2017-08-19
584 text component on time of diary entry

### LINKED ISSUE #584: BUG: TextComponent#on_time? checks system time not diary_entry#moment

roschaefer:
close #584


app/models/text_component.rb
```diff
@@ -20,16 +20,16 @@ class TextComponent < ActiveRecord::Base
   enum publication_status: { :draft => 0, :fact_checked => 1, :published => 2 }
 
   def active?(diary_entry = nil)
```
tillprochaska: `diary_entry`shouldn’t be optional

PR #589, opened 2017-08-19
588 text components have hours constraint

### LINKED ISSUE #588: TextComponents: Move hours constraint from Triggers to TextComponent

roschaefer:
close #588

### DIFF
--- a/app/models/text_component.rb
+++ b/app/models/text_component.rb
@@ -7,6 +7,11 @@ class TextComponent < ActiveRecord::Base
       '(18:00 - 23:00) evenings' => [18, 23].to_json,
       '(23:00 - 06:00) nights' => [23, 6].to_json,
   }
+  # This method associates the attribute ":image" with a file attachment
+  has_attached_file :image, styles: {
+    small: '620>',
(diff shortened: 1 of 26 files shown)


app/models/text_component.rb
```diff
@@ -19,21 +31,40 @@ class TextComponent < ActiveRecord::Base
 
   enum publication_status: { :draft => 0, :fact_checked => 1, :published => 2 }
 
-  def active?(diary_entry = nil)
+  def timeframe=(frame)
```
tillprochaska: That’s something I learnt today! 👍

app/views/text_components/_form.html.haml
```diff
@@ -49,6 +49,9 @@
                 = f.input :from_day
               .col
                 = f.input :to_day
+            .row
+              .col
```
tillprochaska: The `.row` and `.col` wrapping divs are completely unnecessary in this case. You only need them if you want to create a grid with two or more columns like in the example above. In this case they only add bloat and make the markup harder to read.

PR #623, opened 2017-08-24
Enhanced responsive styles

tillprochaska:


tillprochaska:
The breadcrumb nav now collapses on mobile screens. It’s a bit cluttered, though. @drjakob, I know you really want those small messages in the secondary navigation, but it takes a pretty large amount of whitespace. Maybe we could hide them on mobile screens?

img_3075 2

img_3076

PR #625, opened 2017-08-27
553 Remove character limit for questions

tillprochaska:


spec/models/question_answer_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 RSpec.describe QuestionAnswer, type: :model do
   describe '#question' do
-    context 'more than 20 characters' do
-      subject { build(:question_answer, question: ("q" * 21)) }
+    context 'more than 640 characters' do
```
tillprochaska: @roschaefer: Out of curiosity: What’s the reason for the 640 characters limit for questions and answers?

drjakob: Facebook

tillprochaska: Ok. So should the character limit be removed completely for both, questions and answers, or should we leave it at 640 characters?

drjakob: Should be completly  removed. (And what happens to our beautiful character counter btw?)

tillprochaska: The character counter was only shown for the main part of the sensor story text. But if you’d like, I could easily add it to questions and/or answers as well.

drjakob: yes, please

Issue #374, opened 2017-05-30
Green Flashlight for active tcs


drjakob:
As reporter
I want to see a "green flashlight" in text component list marking active tcs
in order to better debug the live text.

sebroeder:
Question related to the logic of `TextComponent#active?`, not the visualization in the UI: A text component is considered active, when it does not have any triggers attached to it ([Implementation]( https://github.com/roschaefer/story.board/blob/09f4d548b8325195b0501f52b609679ee4942135/app/models/text_component.rb#L22-L24), [Spec](https://github.com/roschaefer/story.board/blob/09f4d548b8325195b0501f52b609679ee4942135/spec/models/text_component_spec.rb#L48-L52)).

Is this what we want? It seems a little counter-intuitive to me, but this could of cause just be a misunderstanding on my side.

drjakob:
No, a text component is considered active when its selectet to be (potentially) used in the current sensorstory. 
1. The connected trigger is active 
2. Time span of tc is true 
So the function should have debugging purposes: As an editor I read my sensorstory and want to debug the used tcs.)

sebroeder:
What I wanted to say is that in the current implementation in production (see links in the original comment), a text component is **active** when it does not have any triggers attached.

You can try this yourself in the `bin/rails console`:

```
irb(main):001:0> t = TextComponent.new
=> #<TextComponent id: nil, heading: nil, introduction: nil, main_part: nil, closing: nil, from_day: nil, to_day: nil, report_id: nil, topic_id: nil, assignee_id: nil, publication_status: "draft", created_at: nil, updated_at: nil, notes: nil>
irb(main):002:0> t.triggers
=> #<ActiveRecord::Associations::CollectionProxy []>
irb(main):003:0> t.active?
=> true
```

Would you want it to be **inactive** when it does not have any triggers attached instead? Than we have to change the current implementation and maybe document the desired behavior in a Cucumber feature instead of just the model spec.

Again, this is technically not related to the visualization of the active/inactive status that is the scope of this issue, but a question/surprise that came up when I read the implementation of `TextComponent#active?` in the model.

roschaefer:
It's intended that a text component is active if it has no triggers. E.g. there is a condition on the text component itself, a start and end day. If we would change the behaviour to force text components to have at least one active trigger, we would have to catch this situation.

sebroeder:
[...]

PR #647, opened 2017-09-02
606 show timeframe

### LINKED ISSUE #606: Show timeframe in tc overview
As an editor
i want to see the timeframe in tc overview-list (as a symbol?)
in order to faster get an overview.

tillprochaska:
The timeframe will now be displayed in the text components listing as well.

@matthib: I know this [regex replacement](https://github.com/roschaefer/story.board/blob/606_show_timeframe/app/views/text_components/_table_row.html.haml#L14) is quite dirty, I didn’t want to refactor the[ part of the model](https://github.com/roschaefer/story.board/blob/606_show_timeframe/app/models/text_component.rb#L2-L9) where those labels are generated…

@drjakob: Unfortunately I didn’t find a good way to indicate the timeframe using icons or anything else, but I think simple text is actually a good and unambigous solution in this case.

bildschirmfoto 2017-09-02 um 17 05 54

close #606

### DIFF
--- a/app/assets/stylesheets/components/_item-table.scss
+++ b/app/assets/stylesheets/components/_item-table.scss
@@ -26,6 +26,10 @@
     line-height: 1rem * $line-height-base;
 }
 
+.item-table__meta {
+    white-space: nowrap;
+}
+
 .item-table__actions {
     white-space: nowrap;
 
--- a/app/controllers/text_components_controller.rb
+++ b/app/controllers/text_components_controller.rb
@@ -81,7 +81,7 @@ class TextComponentsController < ApplicationController
 
       filter_text_components
 
-      @trigger_groups = @text_components.order('from_day').group_by(&:trigger_ids)
+      @trigger_groups = @text_components.order('from_day, from_hour').group_by(&:trigger_ids)
 
       @trigger_groups = @trigger_groups.map do |trigger_ids, components|
         [Trigger.find(trigger_ids), components]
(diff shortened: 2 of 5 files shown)


drjakob:
Ok, danke

tillprochaska:
@matthib: Thanks for the review!

Adding a helper method to a decorator or to the model directly that returns either a short or a detailed label for the timeframe would probably clean up things a bit. But I’ll leave it as is for now.

For the second part: Yep, and I think there already is an issue which is pretty similar to your idea: #374. We’ve talked about this in our last phone conference and there where worries that checking the current status for every text component might cause performance issues. Highlighting text components based only on the current time and day should required less database queries etc., but might also be confusing. @drjakob, any thoughts on this? But this is probably  a separate user story anyways.

matthib:
@tillprochaska Yes you are right, I remember talking about that the check for "green light" for every text component would cause a big performance impact. But maybe displaying the current timeframe and day would help as an orientation which text component is active at which point in time. Otherwise you always have to look up what the exact duration of the timeframes is and also know which day the project is currently on. I don't know if I'm the only one getting confused by this, but I guess I would have to double check all the time, to make sure I've set the day and the timeframe correctly.

tillprochaska:
I’m totally with you. My preferred solution would still be the active flash light (which is actually very easy to implement, so we could just try this and see if it actually has a relevant impact on the performance). Displaying the current day since the start date of the project as well as the current time frame seems to be a good alternative to me.

Let’s hear what Jakob or the editors think about this. I think this isn’t crucial for the project start so we could implement this later… ;)

PR #654, opened 2017-09-03
Fix #639

### LINKED ISSUE #639: BUG Alert-Trigger gets lost
BUG: This is a serious bug. 
The Alert (calving will start in 6 hours, sensor (ALERT_Smaxtec_Kalbung_Emma:KuhNr.206_HofHoeck", ID 111), triggered by "Emma Alarm: Geburtsalarm hat ausgelöst. Geburt ist in 6 bis 12 Stunden erwartet"  has been displayed yesterday, 16'00 on the page of Emma. Now its gone. WHY? active Triggers should be stable in history! Otherwise the scroll-back-function has no value if important things gets lost.

tillprochaska:
For details, see the comments for issue #639.

By the way: I was a bit surprised the tests in `reports/time_travel.feature` did not fail. Maybe we should extend these tests to use the validity period as well.

close #639

### DIFF
--- a/app/models/trigger.rb
+++ b/app/models/trigger.rb
@@ -26,7 +26,7 @@ class Trigger < ActiveRecord::Base
         active = true
         active &= (condition.from.nil? || condition.from <= reading.calibrated_value)
         active &= (condition.to.nil? ||reading.calibrated_value <= condition.to)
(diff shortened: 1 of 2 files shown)


drjakob:
YAY!

PR #657, opened 2017-09-03
634 highlight sensors manual input

tillprochaska:


app/controllers/sensors_controller.rb
```diff
@@ -3,7 +3,7 @@ class SensorsController < ApplicationController
   before_action :set_readings, only: [:show, :start_calibration, :stop_calibration]
 
   def index
-    @sensors = Sensor.where(report: @report)
+    @sensors = Sensor.where(report: @report).sort_by{|s| s.manual? ? 0 : 1}
```
roschaefer: is `manual` an attribute in the db? Use `Sensor.where(...).order(:manual)`

app/models/sensor.rb
```diff
@@ -10,6 +10,7 @@ class Sensor < ActiveRecord::Base
   delegate :max, to: :sensor_type
   delegate :unit, to: :sensor_type
   delegate :fractionDigits, to: :sensor_type
+  delegate :data_collection_method, :manual?, :active?, to: :sensor_type
```
roschaefer: I see, but then you can do `Sensor.includes(:sensor_type).order('"sensor_types"."manual"')`..

roschaefer: or `joins` instead of `includes` maybe

app/models/sensor_type.rb
```diff
@@ -1,4 +1,5 @@
 class SensorType < ActiveRecord::Base
   default_scope { order(:property) }
   validates :unit, length: {minimum:  0, allow_nil: false}
+  enum data_collection_method: {automatic: 0, manual: 1}
```
roschaefer: alright :laughing: `data_collection_method` is the actual attribute.. still you can do the ordering with sql

PR #674, opened 2017-09-06
374 green flashlight

tillprochaska:
@drjakob requested this one again and I gave it a quick try. While it certainly needs more db queries to highlight text components that are currently active, I managed to do some optimizations. On my local machine, the perceived loading time didn’t increase too badly…

// cc @drjakob

### DIFF
--- a/app/controllers/text_components_controller.rb
+++ b/app/controllers/text_components_controller.rb
@@ -84,14 +84,14 @@ class TextComponentsController < ApplicationController
     def set_index_data
       @diary_entry = DiaryEntry.new(report: @report, release: :final, moment: Time.zone.now)
 
-      @text_components = TextComponent.includes(:triggers, :question_answers, :channels)
+      @text_components = TextComponent.includes(:report, :channels, :triggers)
 
       filter_text_components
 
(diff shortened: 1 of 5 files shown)


app/controllers/text_components_controller.rb
```diff
@@ -82,14 +82,16 @@ def set_form_data
...
+
+      @text_components = TextComponent.includes(:report, :channels, :triggers)
 
       filter_text_components
 
       @trigger_groups = @text_components.order('from_day, from_hour').group_by(&:trigger_ids)
 
       @trigger_groups = @trigger_groups.map do |trigger_ids, components|
-        [Trigger.find(trigger_ids), components]
+        [components.first.triggers, components]
```
roschaefer: I don't get that one

tillprochaska: I’m not sure if this actually has any advantages, as db requests are automatically cached by active record, still I thought simply using the triggers already loaded might be better than finding them again by id.

roschaefer: hm, I'm not quite sure if this is the same thing - but actually, yeah

PR #680, opened 2017-09-09
Change evening timeframe from 18-0h, change nights timeframe to 0-6h

### LINKED ISSUE #677: CONFLICT: night timeframe conflicting in planing with days
There is a conflict with timeframe night (23-6 o'clock) and days.

As an editor 
planing to publish a tc in the night of day 3 to 4 I choose timeframe night (23-6) and day 3 to four.
BUT then tc will also be active from 23-23:59) on day 4.
Possible Solution: changeing night-timeframe to 0-6 
 
What do you think?

matthib:
Change evening timeframe from 18-0h, change nights timeframe to 0-6h as proposed by @drjakob in #677 

@tillprochaska Do you know if changing the timeframes in the text_component model is sufficient, or do I need to make more changes for the new timeframes to get triggered properly?

--- 

Close #677

### DIFF
--- a/app/models/text_component.rb
+++ b/app/models/text_component.rb
@@ -4,8 +4,8 @@ class TextComponent < ActiveRecord::Base
       '(06:00 - 09:00) in the morning' => [6, 9].to_json,
       '(09:00 - 13:00) before noon' => [9, 13].to_json,
       '(13:00 - 18:00) afternoons' => [13, 18].to_json,
-      '(18:00 - 23:00) evenings' => [18, 23].to_json,
-      '(23:00 - 06:00) nights' => [23, 6].to_json,
+      '(18:00 - 00:00) evenings' => [18, 0].to_json,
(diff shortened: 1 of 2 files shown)


tillprochaska:
Hey @matthib: I haven’t had a closer look at this, but I think we need a rake task to update existing text components. (It’s not the timeframe that’s saved to the db but the actual from and to hours the timeframe is mapping to.)

drjakob:
Rake task sound plausible to me.

tillprochaska:
Yep, that's probably the best way!

matthib:
@tillprochaska Yep, you're right, the text_components from_hour and to_hour needs to be changed for the concerning text_components. 
With the last commit I add a rake task to do exactly that.

roschaefer:
A `migration` would have been the better choice here. Why? We need to change all the data, no matter if `development`, `test` (hopefully empty, anyways) or `production`.

PR #686, opened 2017-09-10
683 validity period change to float

### LINKED ISSUE #683: Validity period should accept 0.5
As an editor
i want Trigger Validity period to accept .5 and 0.25 as values 
in order to time shorter alerts, eg. for drinking status in cows.

matthib:
For the validity period to accept floating numbers it needs to be changed to a float field in the db. I also added a cucumber test to verity that the validity period accepts float numbers.

---

Close #683

### DIFF
--- a/db/migrate/20170910091746_change_validity_period_to_float_in_triggers.rb
+++ b/db/migrate/20170910091746_change_validity_period_to_float_in_triggers.rb
@@ -1,0 +1,5 @@
+class ChangeValidityPeriodToFloatInTriggers < ActiveRecord::Migration[5.0]
+  def change
+    change_column :triggers, :validity_period, :float
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20170910091746_change_validity_period_to_float_in_triggers.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeValidityPeriodToFloatInTriggers < ActiveRecord::Migration[5.0]
+  def change
+    change_column :triggers, :validity_period, :float
```
roschaefer: if you provide the former data type, this migration is reversible

matthib: @roschaefer I couldn't find an example for that, could you provide one please?

PR #713, opened 2017-09-17
615 duplicate text components

### LINKED ISSUE #615: Duplicate Text Components
As editor
I want to easily duplicate text components in same report, heading added "Copy"
in order to easily adapt text components in same or different reports.

tillprochaska:
tc_duplicate 1

close #615 / cc @drjakob

### DIFF
--- a/app/controllers/text_components_controller.rb
+++ b/app/controllers/text_components_controller.rb
@@ -141,7 +141,7 @@ class TextComponentsController < ApplicationController
       params.require(:text_component)
         .permit(:heading, :introduction, :main_part, :closing, :from_day, :to_day,
                 :timeframe, :to_hour, :report_id, :topic_id, :assignee_id, :image, :image_alt, :delete_image,
-                :publication_status, :notes, :duplicate, trigger_ids: [], channel_ids: [],
+                :publication_status, :notes, trigger_ids: [], channel_ids: [],
                 question_answers_attributes: [:id, :question, :answer, :_destroy],
                 triggers_attributes: [:name, :from_hour, :to_hour,
                                       :priority, :report_id,
--- a/app/controllers/text_components_controller.rb
+++ b/app/controllers/text_components_controller.rb
@@ -27,6 +27,7 @@ class TextComponentsController < ApplicationController
     @text_component.report = @report
     @text_component.heading = [master.heading, 'COPY'].join(' ')
     @text_component.publication_status = :draft
+    @text_component.channels = master.channels
     @text_component.triggers.build
 
     master.question_answers.each do |qa|
--- a/app/controllers/text_components_controller.rb
+++ b/app/controllers/text_components_controller.rb
@@ -16,8 +16,24 @@ class TextComponentsController < ApplicationController
 
   def new
     @text_component = TextComponent.new
+    @text_component.report = @report
     @text_component.triggers.build
-    @text_component.report = Report.current
+  end
+
+  def duplicate
+    master = TextComponent.find(params[:id])
+
+    @text_component = master.dup
+    @text_component.report = @report
+    @text_component.heading = [master.heading, 'COPY'].join(' ')
+    @text_component.publication_status = :draft
+    @text_component.triggers.build
+
+    master.question_answers.each do |qa|
+      @text_component.question_answers << qa.dup
+    end
+
+    render :new
   end
 
   def edit
@@ -120,7 +136,7 @@ class TextComponentsController < ApplicationController
         .permit(:heading, :introduction, :main_part, :closing, :from_day, :to_day,
                 :timeframe, :to_hour, :report_id, :topic_id, :assignee_id, :image, :image_alt, :delete_image,
 
-                :publication_status, :notes, trigger_ids: [], channel_ids: [],
+                :publication_status, :notes, :duplicate, trigger_ids: [], channel_ids: [],
                 question_answers_attributes: [:id, :question, :answer, :_destroy],
                 triggers_attributes: [:name, :from_hour, :to_hour,
                                       :priority, :report_id,
--- a/app/views/text_components/_form.html.haml
+++ b/app/views/text_components/_form.html.haml
@@ -128,7 +128,7 @@
             .step__action
               %a.form__section__action--open.btn.btn-sm.btn-secondary{href: '#', 'data-form-toggle': 'open-section'} Edit
         .form__section__body
-          = f.association :channels, selected: @text_component.channels.count > 0 ? @text_component.channel_ids : Channel.find_by({name: 'sensorstory'}).id, input_html: {'data-choices': true}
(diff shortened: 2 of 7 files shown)


drjakob:
Looks cool.

app/views/text_components/_form.html.haml
```diff
@@ -128,7 +128,7 @@
             .step__action
               %a.form__section__action--open.btn.btn-sm.btn-secondary{href: '#', 'data-form-toggle': 'open-section'} Edit
         .form__section__body
-          = f.association :channels, selected: @text_component.channels.count > 0 ? @text_component.channel_ids : Channel.find_by({name: 'sensorstory'}).id, input_html: {'data-choices': true}
+          = f.association :channels, selected: @text_component.channels.length > 0 ? @text_component.channel_ids : 1, input_html: {'data-choices': true}
```
roschaefer: no magic numbers please, use a constant in `Channel`

config/routes.rb
```diff
@@ -30,7 +30,9 @@
   resources :reports, param: :report_id # shallow nesting to avoid id param like :report_report_id
   resources :reports, only: [] do # only: [] to remove duplicate top level routes
     resources :diary_entries, only: [:index, :show]
-    resources :text_components
+    resources :text_components do
+      get 'duplicate', on: :member
```
roschaefer: I believe `,on: :member` is unnecessary (default)

spec/controllers/text_components_controller_spec.rb
```diff
@@ -68,6 +68,41 @@
...
+        expect(assigns(:text_component).question_answers[i].question).to eq(qa.question)
+        expect(assigns(:text_component).question_answers[i].answer).to eq(qa.answer)
+      end
+    end
+
+    it "copys the text component's channels" do
+      expect(assigns(:text_component).channels).to match_array(text_component.channels)
+    end
+  end
+
```
roschaefer: Great! I would also love to see a test that includes a `POST` request. Maybe a cucumber test. I would like to make sure that the entire `TextComponent` and it's association is untouched by the duplicate.

app/controllers/text_components_controller.rb
```diff
@@ -124,7 +141,7 @@ def text_component_params
       params.require(:text_component)
         .permit(:heading, :introduction, :main_part, :closing, :from_day, :to_day,
                 :timeframe, :to_hour, :report_id, :topic_id, :assignee_id, :image, :image_alt, :delete_image,
-                :publication_status, :notes, trigger_ids: [], channel_ids: [],
+                :publication_status, :notes, :duplicate, trigger_ids: [], channel_ids: [],
```
roschaefer: You probably don't need to include the `:duplicate` here as it is not an attribute that will go into the `TextComponent`.

spec/controllers/text_components_controller_spec.rb
tillprochaska: Hm, okay. The duplicate action renders the new view which in turn posts to the new action, so I thought this wouldn’t be necessary. The `dup` method duplicates a record without it’s associations so any associations need to be set explicitly. But you’re right, I could probably add a test to make sure that questions and answers and triggers will always be duplicated or, respectively, removed.

config/routes.rb
tillprochaska: No, it’s not. Sorry to disappoint you… 😿

spec/controllers/text_components_controller_spec.rb
roschaefer: I don't know this behaviour. Just write some tests to show it, so other developers can see the desired, explicit behavior. If it is very low-level, a model test (unit test) might be enough already

config/routes.rb
roschaefer: how about
```ruby
member :text_components do
  get 'duplicate'
end
```
?

roschaefer: it's good either way - just leave it like that

app/controllers/text_components_controller.rb
tillprochaska: Oh, that was actually leftover from my initial approach passing an existing text component’s id to the `new` action…

spec/controllers/text_components_controller_spec.rb
tillprochaska: Okay, I extended the controller test to make sure that the associations are duplicated as well. I started writing a cucumber test as well, but had to deal with timing issues I wasn’t able to fix in a reasonable time. What do you think, are these tests sufficient.

spec/controllers/text_components_controller_spec.rb
```diff
@@ -91,8 +91,13 @@
       expect(assigns(:text_component).draft?).to be_truthy
     end
 
+    it "removes all associated triggers" do
+      expect(assigns(:text_component).triggers.count == 0)
```
roschaefer: good

spec/controllers/text_components_controller_spec.rb
```diff
@@ -91,8 +91,13 @@
...
+    end
+
     it "duplicates questions and answers as well" do
       text_component.question_answers.each.with_index do |qa, i|
+        expect(assigns(:text_component).question_answers[i]).not_to eq(qa)
```
roschaefer: I'm pretty sure, this will catch nothing. Ruby has four equality matchers (`==`, `===`, `equal?`, and `eql?`), none of which compares `Active::Record` in a way you would expect here. I would compare the `id` here. That's an integer that must be different.

tillprochaska: According to the docs it should work: http://api.rubyonrails.org/classes/ActiveRecord/Core.html#method-i-3D-3D

tillprochaska:
> The remedy for all race conditions in cucumber tests is to wait for something that you should see on the page when the request has finished.

That’s exactly what I did. Didn’t solve the issue’s though and seems to be a bit more tricky – once again.

spec/controllers/text_components_controller_spec.rb
roschaefer: Did you try it one time? When I wrote tests like that and I introduced a bug, the rspec never failed. Maybe I can learn sth. here, because I was using the wrong methods.

roschaefer:
if you want, you can push the failing cucumber test and we can have a look on it together

spec/controllers/text_components_controller_spec.rb
tillprochaska: I tried this, but only in the rails console. Will have a closer look tomorrow.

```
irb(main):013:0> tc1 = TextComponent.first
=> #<TextComponent id: 2, …>

irb(main):014:0> tc2 = tc1.dup
=> #<TextComponent id: nil, …>

irb(main):015:0> tc1 == tc1
=> true

irb(main):016:0> tc1 === tc1
=> true

irb(main):017:0> tc1 == tc2
=> false

irb(main):018:0> tc1 === tc2
=> false
```

